PR #11, opened 2020-04-12
Recent fixes and adding the new web app

ryanabooth:
See commits.


ryanabooth:
@maniramezan should we merge this?

maniramezan:
@ryanabooth From `dev` to `master`? If we're ok with having a broken `master` for a while until all pages are done and integrated with the backend on the mobile app, sure. Otherwise, might make sense to cherry pick and only merge web stuff. If we go with this approach, might make sense to retire `dev` branch and keep pushing to `master` until the first official release cut where we introduce back `develop` branch and leave the `master` to be always the last stable release.

PR #22, opened 2020-05-06
Move all app related files to mobileapp

maniramezan:
No new code changes, just moving folders and files around.

> This change is for cleaning up the root folder and making it easier to navigate between web and mobile applications. Also, this helps for anybody who views that hierarchy on the repo, finds out that there are both web and mobile apps in this repo.


ryanabooth:
Nice work. 👍

PR #40, opened 2020-08-19
Add game grid and game sorting

ryanabooth:
https://app.asana.com/0/1161945053441353/1178760399401475/f

### DIFF
--- a/web/src/components/Games.jsx
+++ b/web/src/components/Games.jsx
@@ -213,7 +213,8 @@ const useStyles = makeStyles(theme => ({
     '&:hover': {
       backgroundColor: 'rgba(0, 0, 0, 0.05)',
       cursor: 'pointer',
-    }
+    },
+    height: '152px',
   },
   gameGrid: {
     display: 'inline-block',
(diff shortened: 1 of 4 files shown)


web/src/components/Games.jsx
```diff
@@ -170,9 +171,9 @@ export default function Games({ loading, games, saveGame, saveQuestion, saveNewG
           </Select>
           <NewGameDialogue open={newGameOpen} onClose={() => setNewGameOpen(false)} submit={handleNewGame} />
         </Box>
-        {renderGames()}
+        {renderGames(loading)}
       </Grid>
-      {match && (
```
ryanabooth: Only render the Game form when a game exists at the index set in the URL. This was throwing an error before because the displayed game is chosen by its index in the game array, and that array is now affected (filtered) by the search bar.

web/src/components/Games.jsx
```diff
@@ -73,7 +73,8 @@ export default function Games({ loading, games, saveGame, saveQuestion, saveNewG
     setSortType(event.target.value);
   };
 
-  const renderGames = () => {
+  const renderGames = (loading) => {
+    if (loading) return <Typography gutterBottom>Loading...</Typography>;
```
ryanabooth, line 77: Pass the loading boolean into the game render method, and show a loading message rather than the "No results found." message while the games are loading.

PR #78, opened 2021-01-15
Connect teachers game list to existing data

EricNguyen-B:
Added json file that has existing data to the mobile app project and implemented function to fetch that data


maniramezan (approved):
LGTM, great job : ) One minor note is the common practice for commit message tense is present or present continuous.

Issue #61, opened 2020-11-25
[Mar 10] Create DynamoDB schema for GameState


ryanabooth:
WIP https://github.com/rightoneducation/righton-app/pull/85

ryanabooth:
@maniramezan posted a WIP https://github.com/rightoneducation/righton-app/pull/85

PR #85, opened 2021-02-09
[April 16] ERD and entity analysis for game session entities

maniramezan:
The initial work to implement and model dynamodb scheme. 

The ERD diagram and more details on how this maps to dynamodb comes after initial discussion and approval.

This is tracking by #61


docs/mobile/game-session/game_session_db_design.md
```diff
@@ -0,0 +1,53 @@
+# DynamoDB Design
+
+## Entities
+
+### Config
+
+This entity contains the global settings and configurable parts of the application. The list of items included here is as follow
+
+- Points granted to each team estimated the right popular trick answer
+- Points granted per to teams guessed the right answer
```
ryanabooth, line 10:
```suggestion
- Points granted to each team for guessing the most popular trick answer
- Points granted to teams for guessing the right answer
```

docs/mobile/game-session/game_session_db_design.md
```diff
@@ -0,0 +1,53 @@
...
+- Teams: An array of team entities that involved in the game
+- StartedDateTimeUTC: Date time of when the game started in UTC
+- MultiDeviceMode: Boolean indicating if it's a multi-device with facilitator or not
+
+### Team
+
+- TeamNo / TeamId: Numeric unique number assigned to the team
+- Team name: Name chosen to show fo the team. This helps for future substituting `team 1` string with chosen team names
+- Question: The chosen question for the team
+- GameCode: A 4-digit code to enter to join the team
```
ryanabooth, line 26: We use codes to join teams? I thought there was one code for the whole game.

docs/mobile/game-session/game_session_db_design.md
```diff
@@ -0,0 +1,53 @@
...
+- Points granted per to teams guessed the right answer
+
+### GameSession
+
+The higher level entity contains all information related to a game from start to finish.
+
+- Game: A game object that comes from Aurora. It's a snapshot of the data when the game starts
+- Teams: An array of team entities that involved in the game
+- StartedDateTimeUTC: Date time of when the game started in UTC
+- MultiDeviceMode: Boolean indicating if it's a multi-device with facilitator or not
```
sincwu, line 19: Maybe: "Boolean indicating multi-device scenario" (if True, then one device among a team's multiple devices will be randomly assigned the Facilitator; if False, then there is only one device per team and that device will also have Facilitator privileges)

docs/mobile/game-session/game_session_db_design.md
sincwu: Though "No" is indeed a common abbreviation for Number, might TeamNum or something along those lines be worth considering?

docs/mobile/game-session/game_session_db_design.md
```diff
@@ -0,0 +1,53 @@
...
+- AnswerId: A unique id for each answer
+- Text: The answer chosen by a team member
+- TeamMember: DeviceId entered the answer
+- IsTrickAnswer: Indicates if this answer is chosen to be a trick answer
+
+### TeamRoundResults
+
+- TeamId: The unique number identifies the team choosing the answer
+- AnswerId: Identifies which answer
+- QuestionTeamID: Identifies which team's question is this
```
sincwu, line 53: Might it make sense to further differentiate TeamId from QuestionTeamId -- e.g., perhaps updating the former to AnswerTeamId? For consistency, perhaps QuestionTeamID might be tweaked to QuestionTeamId.

docs/mobile/game-session/game_session_db_design.md
sincwu: Variables configured on the teacher side that affect student game sessions include (variable names are just draft suggestions):
- VotingTime: Amount of time that voting teams have to vote (this equals the amount of time that the team coming up with the trick answers (aka QuestionTeam?) has to come up with and select their trick answers.
- HintsRevealTime: Amount of time that passes before hints are revealed. This must be less than VotingTime.
- We should set default values for both these variables and allow teachers to adjust them before a game starts. VotingTime should be something that a teacher can override in the middle of game (e.g., if a teacher sees that all teams are ready to proceed, the teacher should be able to move on to the results screens without having to wait for the entire VotingTime to have passed).

docs/mobile/game-session/game_session_db_design.md
EricNguyen-B: Following up with Sinclair's comment, it may make consistent sense to update AnswerId to AnswerTeamId as well as the formers if we're looking to distinguish them. Otherwise, I think it makes sense design wise.

docs/mobile/game-session/game_session_db_design.md
```diff
@@ -0,0 +1,53 @@
...
+- Question: The chosen question for the team
+- GameCode: A 4-digit code to enter to join the team
+
+### TeamMember
+
+- TeamId: The teamId this member is a member of
+- DeviceId: Unique number associated to the device
+- IsFacilitator: A boolean indicating if this device is the facilitator in multi-device mode
+
+### Question
```
ryanabooth, line 34: These Question entities would also be in the GameSession.Game. Is this section describing the field types for Questions, or are we thinking about pulling them out of the Game object? If so, does the data ever change?

docs/mobile/game-session/game_session_db_design.md
```diff
@@ -0,0 +1,53 @@
...
+- QuestionId: Unique number for each question
+- Question: The text for the question
+- Hints: an array of hints for the question
+- ImageURL: URL of the image associated to this question
+- Answer: The right answer for this question
+
+### TeamAnswer
+
+- AnswerId: A unique id for each answer
+- Text: The answer chosen by a team member
```
ryanabooth, line 45: How do I compare this to the related question? Should we add a reference to the question here?

docs/mobile/game-session/game_session_db_design.md
```diff
@@ -0,0 +1,53 @@
...
+- TeamMember: DeviceId entered the answer
+- IsTrickAnswer: Indicates if this answer is chosen to be a trick answer
+
+### TeamRoundResults
+
+- TeamId: The unique number identifies the team choosing the answer
+- AnswerId: Identifies which answer
```
ryanabooth, line 52: is this the answer they gave that was correct?

ryanabooth (commented):
I don't see any big missing elements of the existing entities. I'm curious what you think about adding some entity to store the overall score of the game:
image

If not, how would we derive the scoreboard from these data types?

docs/mobile/game-session/game_session_db_design.md
maniramezan: Good point, per our discussion, I added `VotingTime` and `HintsRevealTime` to `Config` also and also here. Regarding `VotingTime` modification at runtime, we should probably just do this through UI so teachers can end it by just tapping a button.

docs/mobile/game-session/game_session_db_design.md
maniramezan: We talked about it, the idea here was assigning codes for each team so audience joins with a code that assigned them to a specific team. For now, that's not needed so moving this to `GameSession`

docs/mobile/game-session/game_session_db_design.md
maniramezan: Yes, actually I missed to layout the `Game` entity here. I'm listing all entities that are involved on the mobile side and then their relationships. Then we can decide if questions will be `GameSession.Game.Questions` or be accessible as a separate entity.

ryanabooth:
To handle the question versions issue, maybe our Question ID in this context could be the Aurora Question ID + the current version number. Ex: `2133124908SFKLJ.4` `{UUID}.{version}`

ryanabooth:
Or the original ID in Aurora could be `{UUID}.{versionCount}`

docs/mobile/game-session/game_session_db_design.md
```diff
@@ -0,0 +1,75 @@
...
+- DeviceId: Unique number associated to the device
+- IsFacilitator: A boolean indicating if this device is the facilitator in multi-device mode
+
+### TeamAnswer
+
+- AnswerId: A unique id for each answer
+- QuestionId: The identifier for the question this answer is for
+- Text: The answer chosen by a team member
+- DeviceId: DeviceId entered the answer
+- IsChosen: Indicates if this answer is chosen to be a trick answer
```
maniramezan, line 66: Change this to `IsChosenTrickAnswer`

PR #107, opened 2021-07-14
[GraphQL] update the schema file

ryanabooth:
@EricNguyen-B Here is an example pull request where I've just copied the schema from AWS and pasted in the new version. There seems to be changes to the order and the number of spaces/tabs, that's why it looks like so many changes. Normally it would be a few lines.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,68 +1,83 @@
-schema {
-  query: Query
-  mutation: Mutation
-}
-
 type Game {
-  cluster: String
-  createdAt: AWSDateTime
-  description: String
-  domain: String
-  grade: String
-  id: ID!
-  standard: String
-  title: String!
-  updatedAt: AWSDateTime
+	id: ID!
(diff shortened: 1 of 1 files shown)


schema.graphql
```diff
@@ -1,68 +1,83 @@
...
+	ReviewingResult
+	Finished
 }
 
 type Mutation {
-  saveGameSession(gameSession: GameSessionInput): GameSession
+	saveGameSession(gameSession: GameSessionInput): GameSession
+	joinGameSession(gameCode: Int!): GameSession!
+	joinGame(gameCode: Int!): GameSession
+	updateGame(gameId: ID!): Game!
```
ryanabooth, line 59:
```suggestion
	updateGame(game: Game!): Game!
```
Because this mutation will edit one or more fields on the game, I think we should pass it an entire Game object.

cc/ @EricNguyen-B @maniramezan

PR #109, opened 2021-09-22
Remove unnecessary AWS config and Amplify references

maniramezan:
Looks like this file is not used in the web and caused troubles to setup our amazing interns to onboard and run the web app : )


EricNguyen-B:
looks good to me!

PR #110, opened 2021-09-22
Fix aws config issue on mobile

ankittrehan2000:
1) Remove the references to AWS config file causing errors during runtime
2) Update `Podfile.lock` to be compatible with latest version of XCode


maniramezan (approved):
Nice job on separating them into two commits. Also thanks for taking care of this for the mobile 🙌

PR #134, opened 2021-11-10
Mobile backend generated from Amplify's CLI

ankittrehan2000:
Created a schema for the AppSync api which will be utilized by the mobile application

### DIFF
--- a/backend/mobile-backend/schema.graphql
+++ b/backend/mobile-backend/schema.graphql
@@ -3,16 +3,17 @@
 input AMPLIFY { globalAuthRule: AuthRule = { allow: public } } # FOR TESTING ONLY!
 
 type GameSession @model {
-  id: ID! @primaryKey(sortKeyFields: ["updatedAt"])
+  id: ID! @primaryKey
   game: Game! @hasOne
   startTime: AWSDateTime
   phaseOneTime: Int
   phaseTwoTime: Int
   teams: [Team] @hasMany
-  updatedAt: AWSDateTime!
+  currentQuestion: Int!
+  isPhaseOne: Boolean!
 }
 
-type Game @model {
+type Game @model(subscriptions: null, timestamps: null) {
   id: ID!
   title: String
   description: String
@@ -23,7 +24,7 @@ type Game @model {
   questions: [Question] @hasMany
 }
 
-type Question @model {
+type Question @model(subscriptions: null, timestamps: null) {
   id: ID!
   question: String!
   answer: String
@@ -33,16 +34,16 @@ type Question @model {
   grade: String
 }
 
-type Team @model(subscriptions: null){
+type Team @model {
   id: ID!
   name: String!
   gameId: ID!
   questionId: ID!
-  trickiestAnswerID: ID!
+  trickiestAnswerIDs: [ID]
   teamMembers: [TeamMember] @hasMany
 }
 
-type TeamMember @model(subscriptions: null) {
+type TeamMember @model {
   id: ID!
   team: Team!
   questionId: ID!
--- a/backend/mobile-backend/schema.graphql
+++ b/backend/mobile-backend/schema.graphql
@@ -1,0 +1,64 @@
+type GameSession @model(subscriptions: null) {
+  id: ID!
+  gameId: ID!
+  game: Game! @connection
+  startTime: AWSDateTime
+  votingTime: Int
+  hintsRevealTime: Int
+  createdAt: AWSDateTime!
+  teams: [Team] @connection(keyName: "byGameSession", fields: ["id"])
+}
+
+type Game @model(subscriptions: null) {
+	id: ID!
+	title: String
+	description: String
+  isAdvanced: Boolean
+  gameCode: Int
+	grade: String
+	questions: [Question] @connection(keyName: "byGame", fields: ["id"])
+}
+
+type Question @model
+@key(name: "byGame", fields: ["gameId", "question"]) {
+	id: ID!
+	gameId: ID!
+  question: String!
+	answer: String
+  trickAnswers: AWSJSON
+	imageUrl: String
+  instructions: AWSJSON
+}
+
+type Team @model(subscriptions: null)
+@key(name: "byGameSession", fields: ["gameSessionId"]) {
+  id: ID!
+  name: String!
+  gameSessionId: ID!
+  gameId: ID!
+  questionId: ID!
+  trickiestAnswerID: ID!
+  teamMembers: [TeamMember] @connection(keyName: "byTeams", fields: ["id"])
+  teamAnswers: [TeamAnswer] @connection(keyName: "byTeams", fields: ["id"])
+  createdAt: AWSDateTime!
+}
+
+type TeamMember @model(subscriptions: null)
+@key(name: "byTeams", fields: ["teamId"]) {
+  id: ID!
+  teamId: ID!
+  team: Team!
+  deviceId: ID!
+  questionId: ID!
+  isFacilitator: Boolean
+}
+
+type TeamAnswer @model
+@key(name: "byTeams", fields: ["teamId"]) {
+  id: ID!
+  teamId: ID!
+  isChosen: Boolean
+  questionId: ID!
+  text: String
+  deviceId: ID!
+}
--- a/backend/mobile-backend/schema.graphql
+++ b/backend/mobile-backend/schema.graphql
@@ -3,23 +3,22 @@
 input AMPLIFY { globalAuthRule: AuthRule = { allow: public } } # FOR TESTING ONLY!
 
 type GameSession @model {
-  id: ID! @primaryKey(sortKeyFields: [updatedAt])
+  id: ID! @primaryKey
   game: Game! @hasOne
   startTime: AWSDateTime
   phaseOneTime: Int
   phaseTwoTime: Int
   teams: [Team] @hasMany
   currentQuestion: Int!
-  isPhaseOne: GameSessionState!
-  updatedAt: AWSTimeStamp!
+  currentState: GameSessionState!
 }
 
 enum GameSessionState {
-	CHOOSINGTRICKANSWER
+	CHOOSING_TRICK_ANSWER
 	FINISHED
-	INITIALINTRO
-	NOTSTARTED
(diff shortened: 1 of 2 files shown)


ankittrehan2000:
Creating a draft because this isn't ready for review yet

ankittrehan2000:
Current schema that I had

Game belongs to GameSession
Game has many Questions
GameSession has many Teams
Team has many Team Members
Team has many Team Answers

ankittrehan2000:
Opening this PR to simplify conversation on the schema

ankittrehan2000:
@maniramezan I just pushed some updates in the latest commit to this schema after looking at the updated properties from the current AppSync schema, looks like it should all be there now

maniramezan (changes requested):
Generally looks great, thanks! Maybe after applying the changes we go through another review and then push it through Amplify to see how it goes

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,58 @@
...
+  id: ID! @primaryKey(sortKeyFields: ["updatedAt"])
+  game: Game! @hasOne
+  startTime: AWSDateTime
+  phaseOneTime: Int
+  phaseTwoTime: Int
+  teams: [Team] @hasMany
+  updatedAt: AWSDateTime!
+}
+
+type Game @model {
```
maniramezan, line 15: IMO `subscription` can be `null` for this and the `Question`

Also for both, we can have `timestamp` to be `null` as well

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,58 @@
...
+  id: ID!
+  question: String!
+  answer: String
+  trickAnswers: AWSJSON
+  imageUrl: String
+  instructions: AWSJSON
+  grade: String
+}
+
+type Team @model(subscriptions: null){
```
maniramezan, line 36: ni: space between ( and }

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,58 @@
...
+  id: ID!
+  name: String!
+  gameId: ID!
+  questionId: ID!
+  trickiestAnswerID: ID!
+  teamMembers: [TeamMember] @hasMany
+}
+
+type TeamMember @model(subscriptions: null) {
```
maniramezan, line 45: I think we'd need to enable `subscription` for this. When game starts, we want to know when each member joins the game

maniramezan (changes requested):
Nice find on the mew changes, happy to see them made into the stable release 👌

ankittrehan2000:
@ryanabooth yeah this is what is currently in progress

backend/mobile-backend/schema.graphql
ankittrehan2000: Yes makes sense! I'll add that subscription in

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,58 @@
...
+input AMPLIFY { globalAuthRule: AuthRule = { allow: public } } # FOR TESTING ONLY!
+
+type GameSession @model {
+  id: ID! @primaryKey(sortKeyFields: ["updatedAt"])
+  game: Game! @hasOne
+  startTime: AWSDateTime
+  phaseOneTime: Int
+  phaseTwoTime: Int
+  teams: [Team] @hasMany
+  updatedAt: AWSDateTime!
```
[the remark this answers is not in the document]
ankittrehan2000: It doesn't like it because of the sortKey. Amplify does not let us define the sort key as updatedAt without it being an explicit field in GameSession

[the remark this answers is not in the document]
ryanabooth: How are we planning to store the current game state? (which question is ongoing, answers, scores)

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,60 @@
+# This "input" configures a global authorization rule to enable public access to
+# all models in this schema. Learn more about authorization rules here: https://docs.amplify.aws/cli/graphql/authorization-rules
+input AMPLIFY { globalAuthRule: AuthRule = { allow: public } } # FOR TESTING ONLY!
+
+type GameSession @model {
```
maniramezan, line 5: I think we can add `updatedAt` and then set the same field in `timestamp` annotation. That way we can have it to be as the sort key and also have amplify handles it for us

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,60 @@
...
+  id: ID! @primaryKey
+  game: Game! @hasOne
+  startTime: AWSDateTime
+  phaseOneTime: Int
+  phaseTwoTime: Int
+  teams: [Team] @hasMany
+  currentQuestion: Int!
+  isPhaseOne: Boolean!
```
maniramezan, line 13: Initially, we had a game status that was an enum to indicate the state of the game. I think we're missing that here. We should add any new state needed to that enum

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,60 @@
...
+  imageUrl: String
+  instructions: AWSJSON
+  grade: String
+}
+
+type Team @model {
+  id: ID!
+  name: String!
+  gameId: ID!
+  questionId: ID!
```
maniramezan, line 41: Probably won't need `gameId` here. We can also have the `questionId` to be actually the question entity

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,60 @@
...
+  trickiestAnswerIDs: [ID]
+  teamMembers: [TeamMember] @hasMany
+  score: Int
```
maniramezan, line 44: We can probably mark this as non-null and add a default value of 0

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,60 @@
...
+}
+
+type TeamMember @model {
+  id: ID!
+  team: Team!
+  questionId: ID!
```
maniramezan, line 50: Since we can use the team.question, we can probably omit this here

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,60 @@
...
+  isFacilitator: Boolean
+  memberAnswers: [TeamAnswer] @hasMany
+}
+
+type TeamAnswer @model {
+  id: ID!
+  isChosen: Boolean
+  text: String
+  deviceId: ID!
```
maniramezan, line 59: We should move this to TeamMember

maniramezan (changes requested):

backend/mobile-backend/schema.graphql
ankittrehan2000: I added the updatedAt with timestamp annotation (pushed a new version, not sure if that is what you meant here) but Amplify doesn't like Sortkeys other than type of string

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,68 @@
...
+
+type GameSession @model {
+  id: ID! @primaryKey(sortKeyFields: [updatedAt])
+  game: Game! @hasOne
+  startTime: AWSDateTime
+  phaseOneTime: Int
+  phaseTwoTime: Int
+  teams: [Team] @hasMany
+  currentQuestion: Int!
+  isPhaseOne: GameSessionState!
```
maniramezan, line 13: Lets probably name this state / currentState or something similar

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,68 @@
...
+  updatedAt: AWSTimeStamp!
```
maniramezan, line 14: We can remove this as `timestamp` will add this by default

backend/mobile-backend/schema.graphql
maniramezan: Huh bummer, lets then see if we can have the state / isPhaseOne field to be the sortkey. If not, We can have that one also to be indexed

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,68 @@
...
+}
+
+enum GameSessionState {
+	CHOOSINGTRICKANSWER
+	FINISHED
+	INITIALINTRO
+	NOTSTARTED
+	REVIEWINGRESULT
+	VOTING
+}
```
maniramezan, line 24: Nit: Lets separate words with `_` to make them more readable

backend/mobile-backend/schema.graphql
```diff
@@ -0,0 +1,68 @@
...
+  grade: String
+}
+
+type Team @model {
+  id: ID!
+  name: String!
+  question: Question! @hasOne
+  trickiestAnswerIDs: [ID]
+  teamMembers: [TeamMember] @hasMany
+  score: Int!
```
maniramezan, line 53: Lets add `@default(value: 0)` directive here

maniramezan (changes requested):

backend/mobile-backend/schema.graphql
ankittrehan2000: So this one doesn't seem to be a directive, all solutions for default values suggest to update the resolvers directly incase if it is missing

ankittrehan2000: Weirdly this one complains about type not being String. Looks like a bug on Amplify CLI
Screen Shot 2022-03-09 at 11 21 02 AM

backend/mobile-backend/schema.graphql
ankittrehan2000: After trying this, got the same complaint about type not being a String. Not quite sure what you mean there by index it

backend/mobile-backend/schema.graphql
maniramezan: Huh, bummer. Feel free to open an issue with Amplify team if you were interested

backend/mobile-backend/schema.graphql
maniramezan: Sounds good. We can look into it later to see what'd be alternatives

maniramezan (approved):
Thanks again for the changes ♥️

PR #160, opened 2022-01-10
navbar for explore, and game maker

RWu603:
Navbar styling with icon created an empty GameMaker file just for reference


RWu603:
image

PR #164, opened 2022-01-17
Modify schema and back up resolvers

ankittrehan2000:
Add updated schema files and back up latest resolvers after question modification to source control


maniramezan (approved):
LGTM ❤️

PR #169, opened 2022-01-18
Add Sort By Dropdown and Fix Filtering

JJackson914:
image
image

- Add sort by dropdown menu onto game dashboard
- Style dropdown to appear like Figma (font family, font sizing, menu option position, etc. still need tweaking)
- Add all menu options from Figma (some options are hidden from view due to not having their filtering added yet)
- Add Oldest filtering and Z to A filtering
- Fix issue with Alphabetical and Z to A filtering treating uppercase and lowercase letters differently. Both filters should now work accurately _most_ of the time.
- Last Updated and Oldest filters still need to be fixed. Once the API is corrected both filters should work correctly automatically.
- Fix some React errors in other files


EricNguyen-B (approved):
This looks good to me!

zach2025 (approved):

Issue #183, opened 2022-02-01
Hide Search Bar


zach2025:
Currently, the Search Bar, located in the Navigation Bar area, displays on all pages. The Search Bar should be hidden on all pages **_expect_** for the Explore/Home page.

**How it should look:**
Explore/Home page
<img width="810" alt="Screen Shot 2022-02-01 at 2 41 45 PM" src="https://user-images.githubusercontent.com/90917774/152040015-660dbe44-c17a-4209-95f3-35eb9bcf9915.png">

Every other page
<img width="810" alt="Screen Shot 2022-02-01 at 2 41 25 PM" src="https://user-images.githubusercontent.com/90917774/152040040-8d342151-18fd-4b1a-ab87-ea6974f0aeaf.png">

PR #187, opened 2022-02-11
Search Bar Update

drewjhart:
Addressing Issue #183: "Hide Search Bar"

Update to ensure search bar is hidden when not on Explore page.

### DIFF
--- a/web/src/components/Nav.jsx
+++ b/web/src/components/Nav.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import {
   NavLink
 } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


web/src/components/Nav.jsx
```diff
@@ -34,22 +33,9 @@ export default function PrimarySearchAppBar({ setSearchInput, searchInput }) {
               </Typography>
             </NavLink>
             <img src={ComingSoon} alt="Coming Soon!!" style={{height: 50, marginLeft: 50, marginRight: 20}} />
+            {matchSearchBar.isExact ? <SearchBar setSearchInput={setSearchInput} searchInput={searchInput} /> : null}
+            {!matchGamesPage.isExact ? console.log("found it") : null}
```
maniramezan, line 37: Is this a debug code that's left here?

maniramezan (changes requested):

web/src/components/Nav.jsx
drewjhart: Thanks for catching that - that's old code. I've pushed a new commit that matches what I showed you on slack

maniramezan (approved):
LGTM, nice job ❤️

zach2025 (approved):
Looks good to me too!

PR #193, opened 2022-02-15
Add authentication from Cognito to web app

ankittrehan2000:
Log In page:

Screen Shot 2022-02-15 at 3 34 10 PM


web/src/aws-exports.js
```diff
@@ -1,13 +0,0 @@
-/* eslint-disable */
-// WARNING: DO NOT EDIT. This file is automatically generated by AWS Amplify. It will be overwritten.
-
-const awsmobile = {
-    "aws_project_region": "us-east-1",
-    "aws_appsync_graphqlEndpoint": "https://rtta7pxyhrdwrlg3d2vrw2mdri.appsync-api.us-east-1.amazonaws.com/graphql",
-    "aws_appsync_region": "us-east-1",
-    "aws_appsync_authenticationType": "API_KEY",
-    "aws_appsync_apiKey": "da2-bnjdiewpkjh2dl7y7dyz4kn7bu"
```
maniramezan, line 9: Nice catch! : )

Would you mind to add this file to the root directory `.gitignore`? Looks like we accidentally merge this in which has some data that shouldn't exist in the git. We might need to do something like [this](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository). But with how there's a sprint1 branch still not merged into `dev`, might cause troubles for others

web/src/components/auth/SignUp.tsx
```diff
@@ -0,0 +1,185 @@
...
+    try {
+      await Auth.signUp({
+        username: email,
+        password: confirmPassword,
+        attributes: {
+          email,
+          name,
+        },
+      });
+      alert("Success");
```
maniramezan, line 43: Might wanna replace these with `console.log` or just remove them before merging

web/src/components/auth/SignUp.tsx
```diff
@@ -0,0 +1,185 @@
...
+      history.push("/confirmation");
+    } catch (error) {
+      console.error(error);
+      alert("Error");
```
maniramezan, line 47: Same as above

web/src/components/auth/SignUp.tsx
```diff
@@ -0,0 +1,185 @@
...
+import TextField from "@material-ui/core/TextField";
+import { styled } from "@material-ui/core/styles";
+import { Auth } from "aws-amplify";
+import Button from "@material-ui/core/Button";
+import CircularProgress from "@material-ui/core/CircularProgress";
+import { Link, useHistory } from "react-router-dom";
+import { Grid } from "@material-ui/core";
+import RightOnLogo from "./RightOnLogo.png";
+
+const DLink = styled(Link)({
```
maniramezan, line 11: I might be missing something here, but doesn't look like this is being used

web/src/components/auth/SignUp.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  return (
+    <Grid
+      container
+      direction="column"
+      alignItems="center"
+      justifyContent="center"
+    >
+      <img
+        src={RightOnLogo}
```
maniramezan, line 60: nit: Not important, but we should probably use a vector based image for the logo, e.i. svg, so it supports different scales doesn't get pixelated

web/src/components/auth/LogIn.tsx
```diff
@@ -0,0 +1,94 @@
...
+      </h1>
+      <Field variant="outlined" label="Email" value={email} type="email" onChange={(e) => setEmail(e.target.value)} />
+      <Field variant="outlined" label="Password" type="password" value={password} onChange={(e) => setPassword(e.target.value)} />
+      <ButtonGrid item direction="row" justifyContent="space-between" spacing={4}>
+      <SignUpLink to="/signup">Sign Up</SignUpLink>
+      <LogInLink to="#" onClick={(e) => {handleSubmit(e)}}>Log In</LogInLink>
+      </ButtonGrid>
+    </form>
+    </Grid>
+    </Grid>
```
maniramezan, line 60: nit: Indent

web/src/components/auth/LogIn.tsx
```diff
@@ -0,0 +1,94 @@
...
+    e.preventDefault();
+    setLoading(true);
+
+    try {
+      await Auth.signIn(email, password);
+      alert("Success!!");
+      //history.push("/");
+      window.location.href="/";
+    } catch (error) {
+      alert("Error!!");
```
maniramezan, line 24: Same for this

web/src/components/auth/Confirmation.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const [code, setCode] = useState("");
+  const history = useHistory();
+
+  const handleSubmit = async (e: React.SyntheticEvent<Element, Event>) => {
+    e.preventDefault();
+    setLoading(true);
+
+    try {
+      await Auth.confirmSignUp(email, String(code));
+      alert("Success");
```
maniramezan, line 23: Same for this and the below alert. Generally, for errors, we might wanna show some message in the UI to tell the user "username or password was wrong" or something like that. Maybe run it by the UX team for what copy to be used

maniramezan (changes requested):

ankittrehan2000:
@maniramezan made some changes here, but will keep this open until the sprint 1 branch is merged in along with @drewjhart's work so that there is isn't any breaking changes on their end. Once they merge in I can deal with with any merge issues (cc: @EricNguyen-B)

maniramezan:
Thanks @ankittrehan2000 , sounds good to me : ) Let me know if I could help with any possible merge conflict or changes come in. I can review it again after the merge

web/src/aws-exports.js
ryanabooth: I added this to git so that the other web developers wouldn't have to generate this locally with amplify. You're welcome to remove it, but we'll need to help them all get this set up locally.

ankittrehan2000: Good point, I can keep it in here then for now

PR #200, opened 2022-02-23
Ray Wu  add question form

RWu603:
Updated the location for add question file.


web/src/components/GameMaker.jsx
```diff
@@ -294,8 +301,24 @@ export default function GameMaker({game, newSave, editSave, gamemakerIndex}) {
 
                 <Grid container item xs={2}></Grid>
             </Grid>
-        </form>
-    );
+
+            <Grid container xs={2}></Grid>
```
zach2025, line 305: I think this was accidentally repeated so this line can be removed since it is already in line 302

web/src/components/AddQuestion.jsx
```diff
@@ -109,30 +109,21 @@ const useStyles = makeStyles(theme => ({
...
-    const newGame = { ...game };
-    const copy = { ...newGame[`q${newIndex}`] };
-    newGame[`q${newIndex}`] = newGame[`q${currentIndex}`];
-    newGame[`q${currentIndex}`] = copy;
-    saveGame(newGame).then(() => history.push('/games/1'));
-    setAnchorEl(null);
-    setActiveIndex(null);
-  };
+  
+  const handleColoneQuestion = async (question) => {
```
zach2025, line 119: nitpick: misspelling

web/src/components/AddQuestion.jsx
```diff
@@ -184,13 +175,13 @@ function AddQuestion({ loading, gameIndex, game, saveGame, deleteQuestion, selec
       })}
       <Grid container className={classes.parent}>
         <Grid item xs={2}>
-          <Button className={classes.greenButton} color="primary" type="button" variant="contained" onClick={() => addQuestion(questions[index])}>Add to Game</Button>
+          <button className={classes.greenButton} color="primary" type="button" variant="contained" onClick={() => handleColoneQuestion(questions[index-1])}>Add to Game</button>
```
zach2025, line 178: nitpick: misspelling (only change if the misspelling of the actual function above on line 119 because those names need to match)

zach2025 (changes requested):
Most of it looks good, just a few little things that need to be changed

zach2025 (approved):

EricNguyen-B (approved):
This looks good to me!

PR #211, opened 2022-03-08
image subfolders

drewjhart:
PR to add subfolders to the image folder. These are organized as per below:
.../src/images:

/icons -> small icons for use on components etc
/logos -> different versions of the righton logo
/narrators -> the monsters
/graphics -> a final catch all folder for misc files (like frames, phone and display images, etc)

File formats: 
jpg -> single scale, no transparency
png -> single scale, transparency
svg -> vector graphics, scalable

Naming:
image-name -> descriptive, separated by hyphens

Ryan also mentioned that these could be placed as near to the component as possible. I don't see any component subfolders right now, so I thought the src folder would be alright. Just let me know if you have any suggestions!


EricNguyen-B:
This looks great Drew! I appreciate the in depth description of the image subfolders and what each subfolder would contain. I agree that the src folder is good for now

EricNguyen-B (approved):

PR #212, opened 2022-03-11
Sample object and API request

ankittrehan2000:


samples/sampleAPIcall.js
```diff
@@ -0,0 +1,8 @@
+import { getGameSession } from "./queries";
+import { API, graphqlOperation } from 'aws-amplify';
+
+export const getGameSession = async () => {
```
ryanabooth, line 4: wouldn't this need an id?

ankittrehan2000: Yes this is my bad, this should have been listGameSessions, but I'll push both to avoid any confusion

JJackson914 (commented):
My page needs a game's CCSS values, as well as a value that keeps track of how many students have joined a game code.

PR #245, opened 2022-05-31
Storybook Implementation for StartGame.jsx

drewjhart:
Initial Storybook implementation for StartGame.jsx using gamesession.json. Stories include "Initial" (0 players), "Three Players" and "Five Players." 

This is still very much in-progress but given that we're trying to do more PR's to keep our workflows closer in synch, I thought I'd put it up


drewjhart:
I've narrowed this down to only affect a few files so I think I'll push this so you can just pull dev to work on StartGame.jsx, Lucah. If anyone sees any issues, just let me know and I'll revert.

PR #253, opened 2022-06-14
Implement live game question card ui

EricNguyen-B:
Implementing the question card ui. Further changes will reflect intended design. 

image

image


drewjhart:
Code looks good and the show more expansion seems really nice when I test it. Might be nice to storybook this so we get a sense of required props when integrating (I think GameInProgress will need to be tweaked to make sure it's passing what this component needs). 

Testing itself (not related to your code but broadly) is a bit of a pain though. Maybe we can talk about it at the meeting today.

PR #257, opened 2022-06-21
created base for phase 2 games storybook

EricNguyen-B:
image

Next steps are to implement non-click on the correct answer and work with integrating the timer header on phase 2

### DIFF
--- a/web/src/host/components/AnswersInProgress.jsx
+++ b/web/src/host/components/AnswersInProgress.jsx
@@ -3,9 +3,7 @@ import { makeStyles } from '@material-ui/core/styles';
 import { Grid, Typography, Card, CardContent, Collapse, IconButton, LinearProgress, Box } from "@material-ui/core";
 import { ExpandMore, Check } from '@material-ui/icons';
 
-
-
-export default function HostAnswerDropdown({answer, explanation, correct, phase2}) {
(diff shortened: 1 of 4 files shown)


EricNguyen-B:
Altered GamesInProgress.jsx to begin combining the QuestionCard, Answers, and Footer. It looks like this:

image

Approached changes to the answers when entering phase2 based on conditional rendering of a boolean value prop "phase2". When phase2 is the boolean true, then it will change stylistically to this:

image

EricNguyen-B:
I'm having difficulty on having two separate stories that look like this.

Phase 1 Answers:

image

Phase 2 Trick Answers:

image

The last answer for both of these screenshots depend on two boolean arguments being correct and phase2.

`<HostAnswerDropdown answer={correctAnswer[0].choice} explanation={correctAnswer[0].explanation} correct={true} phase2={true}/>`

`GamesInProgress.stories.jsx` uses `GamesInProgress.jsx` which will be the page wrapper for all the components that will make up the GamesInProgress page, i.e. Header, QuestionCard, Answers, Footer.  I don't know how to access the argtypes of `HostAnswerDropDown` because it's being wrapped by a parent component named `GameDetails` in `AnswersInProgressDetails.jsx` which has no args.

PR #259, opened 2022-06-27
initiating README.md with running debug panel locally steps

lucahendicott:
Added the steps for getting the debug panel to run locally. Please adjust or edit as you see fit!


maniramezan (approved):
Thanks a lot! ❤️

Issue #263, opened 2022-07-12
RightOn Host: Game In Progress Footer


lucahendicott:
Footer updates

![Image](https://user-images.githubusercontent.com/70814349/190248909-5eb1ff40-31af-42ca-948d-d3f1aa84c7a7.png)

PR #295, opened 2022-08-11
Game In Progress Footer 

drewjhart:
https://github.com/rightoneducation/righton-app/issues/263 per https://docs.google.com/document/d/1m6cThnNOGvm_fJJn1LHGxghtkq4Cv4U5mgAJEqvovTw/edit


drewjhart:
hey @lucahendicott, I think this is ready for review. There is some monkey business on some of the files that I needed to do to get it to run on my end. Here's a list:

1.  `[debug-panel/package.json]` - added && to get it to run on my end
2.  '[networking/package.json]' - added && to get it to run on my end, some other dependencies added that can probably be rejected
3. `[web/src/host/containers/GameSessionContainer.tsx]` - this is the main one. I had to comment out the subscription here to get it to run. I also had to comment out "endpoint;" on line 15 of `networking/lib/src/ApiClient.js` This is related to the same issue. I think Mani fixed this, but I needed to `npm run build` again after adding in the new API call, so the issue returned. I'm guessing we can probably uncomment these lines in our main file. 

I'll post a full description and demo videos in dev chat.

PR #304, opened 2022-08-15
Game In Progress Footer 2

drewjhart:


drewjhart:
ok lucah, I cleaned out all those dependency changes

networking/src/ApiClient.ts
```diff
@@ -90,6 +90,22 @@ export class ApiClient implements IApiClient {
 
     }
 
+    async updateGameSessionFooter(id: string, gameState: GameSessionState, nextQuestion: number, phaseOneTimeReset: number, phaseTwoTimeReset: number): Promise<IGameSession> {
```
maniramezan, line 93: We should name the methods with more generic names. For example, here `updateGameSessionFooter` doesn't mean anything in mobile or any other place other than the Host app. And more, if the position of the information changes, then the method's name loses its meaning in the Host app as well.

In addition, looks like the difference between this and the previous `updateGameSessions` is some extra variables. Instead, we might be able to create one method that handles updating `GameSession` based on the input variables. Otherwise, we'll end up with multiple methods with duplicate logic in them.

maniramezan, line 93: What's the purpose of `phaseTwoTimeReset` and `phaseOneTimeReset`? Is teacher able to change the default values for phase one and phase two timers during the game session?

PR #314, opened 2022-08-17
lucah endicott  - update mapping in host components

lucahendicott:
- Add `title` field to IGameSession so we can access the title of a game for the start game page.
- Update mapping in host components to clear console errors.
- Remove console logs.

### DIFF
--- a/networking/src/Models/IGameSession.ts
+++ b/networking/src/Models/IGameSession.ts
@@ -14,6 +14,7 @@ export interface IGameSession {
     gameCode: number
     currentTimer?: number | null
     questions: Array<IQuestion>
+    title?: string | null
     updatedAt: string
     createdAt: string
 }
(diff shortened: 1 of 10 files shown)


networking/src/Models/IGameSession.ts
```diff
@@ -14,6 +14,7 @@ export interface IGameSession {
     gameCode: number
     currentTimer?: number | null
     questions: Array<IQuestion>
+    title?: string | null
```
maniramezan, line 17: What's `title`? Is this a new field needed for game session?

maniramezan (changes requested):

networking/src/Models/IGameSession.ts
lucahendicott: Copied from game object I believe? I see the title field when I console log the response in the getGameSession API call - but when I call it in the use effect in the GameSessionContainer component I don't get access to the title field. I can see in AWS console that there is a game title field - so I know the object is already set up with it. But yea - couldn't figure out how to access it without adding the field in the IGameSession

PR #315, opened 2022-08-18
Lucah endicott  host question index

lucahendicott:
Had issues with the current question index displaying properly in the header of the `GameInProgress` page. The length of how many questions in the questions array was also being miscalculated - by reading a length of 1 extra question object. 

Add an `if` statement in the `GameInProgress` page to set the `currentQuestionId` to `1` if it is `null`

This does seem to fix the issue of `.length` properly calculating the amount of question objects that are in the array, and also seems to fix the index marker and header in the UI.

### DIFF
--- a/networking/src/ApiClient.ts
+++ b/networking/src/ApiClient.ts
@@ -73,6 +73,7 @@ export class ApiClient implements IApiClient {
 
     async getGameSession(id: string): Promise<IGameSession> {
         let result = await API.graphql(graphqlOperation(getGameSession, { id })) as { data: any }
+        console.log(result)
         return GameSessionParser.gameSessionFromAWSGameSession(result.data.getGameSession)
     }
 
@@ -250,7 +251,9 @@ class GameSessionParser {
             questions,
             currentTimer,
             updatedAt,
-            createdAt
+            createdAt,
+            title
+            
         } = awsGameSession || {}
 
         if (
@@ -281,7 +284,8 @@ class GameSessionParser {
             currentTimer,
             questions: GameSessionParser.mapQuestions(questions.items),
             updatedAt,
-            createdAt
+            createdAt,
+            title
         }
         return gameSession
     }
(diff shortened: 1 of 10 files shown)


networking/src/ApiClient.ts
```diff
@@ -250,7 +250,9 @@ class GameSessionParser {
             questions,
             currentTimer,
             updatedAt,
-            createdAt
+            createdAt,
+            title
+
```
maniramezan, line 255: Nit: extra line

web/src/host/containers/GameSessionContainer.tsx
```diff
@@ -20,20 +20,21 @@ const GameSessionContainer = () => {
...
+    });
 
     //@ts-ignore
     return () => gameSessionSubscription?.unsubscribe()
   }, []);
 
   const handleUpdateGameSessionState = (gameSessionState: GameSessionState) => {
-    apiClient.updateGameSession(gameSessionId, gameSessionState)
-      .then(response => {
+    apiClient.updateGameSession({id: gameSessionId, currentState: gameSessionState})
```
maniramezan, line 36: Oops, missed this one in my updates, thanks for updating it

web/src/host/pages/GameInProgress.jsx
```diff
@@ -13,32 +13,33 @@ export default function GameInProgress({
...
-    teams && teams.items.map(team => 
+    teams && teams.map(team => 
        team.teamMembers && team.teamMembers.items.map(teamMember => 
         teamMember.answers && teamMember.answers.items.map(answer => answer.isChosen && count++
     )))
 
     return count;
   };
 
+  if(currentQuestionId == null) {
```
maniramezan, line 35: Haven't gone through the rest of code, so sorry if it's an obvious one, but why can't we have this to be `null`? Also for adding a default value we should use `questions[0].id` or something

maniramezan (approved):

PR #345, opened 2022-09-23
update map questions in networking 

lucahendicott:
Update `mapQuestions` in `networking` > `ApiClient.ts` to stringify the JSON objects for question choices and instructions to resolve error "Unexpected token 'o', "[object Obj"... is not valid JSON" when attempting to query a game session in both `debug-panel` and `web`

### DIFF
--- a/networking/src/ApiClient.ts
+++ b/networking/src/ApiClient.ts
@@ -1,379 +1,466 @@
-import { IGameSession } from './Models/IGameSession'
-import { IApiClient } from './IApiClient'
+import { IGameSession } from "./Models/IGameSession";
+import { IApiClient } from "./IApiClient";
 import {
-    CreateTeamAnswerInput,
-    CreateTeamAnswerMutation,
-    CreateTeamAnswerMutationVariables,
-    CreateTeamInput,
-    CreateTeamMemberInput,
-    CreateTeamMemberMutation,
-    CreateTeamMemberMutationVariables,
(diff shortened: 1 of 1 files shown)


maniramezan:
Hey, probably need to turn off the semi-colon setting in the VS as it's now showing the whole file as changed and hard to know the real diff

networking/src/ApiClient.ts
```diff
@@ -1,379 +1,466 @@
...
+      .map((awsQuestion) => {
+        if (isNullOrUndefined(awsQuestion)) {
+          throw new Error("Question cannot be null.")
+        }
+        const question: IQuestion = {
+          id: awsQuestion.id,
+          text: awsQuestion.text,
+          choices: isNullOrUndefined(awsQuestion.choices)
+            ? []
+            : JSON.parse(JSON.stringify(awsQuestion.choices)),
```
lucahendicott, line 442: @maniramezan stringify update here

networking/src/ApiClient.ts
```diff
@@ -1,379 +1,466 @@
...
+          imageUrl: awsQuestion.imageUrl,
+          instructions: isNullOrUndefined(awsQuestion.instructions)
+            ? []
+            : JSON.parse(JSON.stringify(awsQuestion.instructions)),
```
lucahendicott, line 446: @maniramezan stringify update here as well

PR #359, opened 2022-10-11
Subscriptions and Other Issues from Integration Test #1

drewjhart:


drewjhart:
Hi all, 
More info + video here: https://righton-education.slack.com/archives/C02D6KZ029L/p1665964590488409

Here's what's been done: 

- subscriptions to CreateTeam, DeleteTeam, CreateTeamAnswer and then passing the responses down into the various StartGame and GameInProgress components. 
- Updated the countdown page to make use of the mobile stuff. reworked it from react-native to reactjs
- fixed the UI issues that came up from last week's testing call
- fixed the page refresh issue by reworking initial usestate and subscriptions

Here's what needs to be done:

- outstanding timer stuff (updating the serverside timer every few seconds per lucah)
- testing with mobile

PR #369, opened 2022-10-19
Lucah endicott  mobile phase 2 basic mode

lucahendicott:
- Initial set up for phase 2 basic mode
- Remove code for advanced mode in basic mode components
- Disable on press for correct answer option
- Update `JoinGame` component with switch case for navigating based on game session `currentState`
- Map wrong answer reasons - TODO: display wrong answer reasons in individual cards properly. 
- TODO: disable all answer options once the user submits an answer


mobile/ios/Podfile.lock
```diff
@@ -381,7 +381,7 @@ PODS:
     - React
   - rn-fetch-blob (0.12.0):
     - React-Core
-  - RNCAsyncStorage (1.17.9):
+  - RNCAsyncStorage (1.17.10):
```
maniramezan, line 384: Why is this updated? Did you updated it in package.json and not pushed?

PR #371, opened 2022-10-22
Host: timer pushes update to gameSession to keep in synch

drewjhart:


drewjhart:
PR to synch timer with `currentTimer` value on AWS. More info can be found here: https://righton-education.slack.com/archives/C02D6KZ029L/p1666547910293609

Here's list of what's been done:

- timers hoisted to GameSessionContainer
- gameTimer added to localstorage in case of page refresh
- timer checks added to gameSession updates 
- UI changes as a result of timer reaching zero 

Here's what still needs to be done:

- testing with mobile for team/answer subscriptions
- testing with mobile for timer synching
- I am getting a eslint issue about missing apiClient and gameSessionId from my 3s timer here:

```
useEffect(()=>{ //update gameSession currentTimer every 3 seconds with time from headerGame Timer 
    if (gameTimer && !gameTimerZero){
      let refreshIntervalId = setInterval(() => {
          let newUpdates = {currentTimer: (localStorage.getItem('currentGameTimeStor')>= 0 ? localStorage.getItem('currentGameTimeStor') : 0)};
          apiClient.updateGameSession({ id: gameSessionId, ...newUpdates });
      }, 3000);
      return () => clearInterval(refreshIntervalId);
    }
  }, [gameTimer, gameTimerZero]);
```

I believe that (for MVP) I can ignore this issue because apiClient and gameSessionId will remain stable for the duration of a game. Is this right though?

networking/src/index.ts
```diff
@@ -1,7 +1,7 @@
 // created from 'create-ts-index'
 
-export * from './AWSMobileApi'
 export * from './ApiClient'
+export * from './AWSMobileApi'
```
maniramezan, line 4: I pushed a commit in `dev` to add this to `.gitignore` list so it doesn't show up. Feel free to delete this file locally and merge the `dev`. This file is auto-generated and no need to be in the source control

host/src/containers/GameSessionContainer.tsx
```diff
@@ -43,7 +48,10 @@ const GameSessionContainer = () => {
 
     let gameSessionSubscription: any | null = null;
     gameSessionSubscription = apiClient.subscribeUpdateGameSession(gameSessionId, response => {
-       setGameSession(({ ...gameSession, ...response }));
+      if (gameSession && gameSession.currentState !== response.currentState) //only run the gametimer check on instances where the currentState changes (new screens)
+        checkGameTimer(response); 
+
+      setGameSession(({ ...gameSession, ...response }));
```
maniramezan, line 54: nit: Probably an extra `()`

host/src/containers/GameSessionContainer.tsx
```diff
@@ -104,20 +112,66 @@ const GameSessionContainer = () => {
     }
   }, [isModalOpen]);
 
+  useEffect(() => { //headerGame timer, 1 second interval, update localstorage in case page resets
```
maniramezan, line 115: nit: The common practice is:
- Having the comments to be in a new line
- A space after `//`

host/src/containers/GameSessionContainer.tsx
```diff
@@ -104,20 +112,66 @@ const GameSessionContainer = () => {
...
+          localStorage.setItem('currentGameTimeStor', headerGameCurrentTime-1);
+        }
+        else
+          setGameTimerZero(true);
+      }, 1000);
+      return () => clearInterval(refreshIntervalId);
+    }
+  }, [gameTimer, gameTimerZero, headerGameCurrentTime]);
+
+  useEffect(()=>{ //update gameSession currentTimer every 3 seconds with time from headerGame Timer
```
maniramezan, line 129: Same nit as above in addition to space before and after `=>`

host/src/containers/GameSessionContainer.tsx
```diff
@@ -104,20 +112,66 @@ const GameSessionContainer = () => {
...
+      }, 3000);
+      return () => clearInterval(refreshIntervalId);
+    }
+  }, [gameTimer, gameTimerZero]);
+
   const handleUpdateGameSession = (newUpdates: Partial<IGameSession>) => {
     apiClient.updateGameSession({ id: gameSessionId, ...newUpdates })
       .then(response => {
+
+        if (response.currentState === stateArray[2])
```
maniramezan, line 143: I'm not familiar with the host app, but this looks to be error-prune as if we change the `GameSessionState` order, it breaks it. Any reason for not using `GameSessionState` straight here? I'd highly recommend against this pattern as it's coupling the order of `GameSessionState` enum values to what they mean in the host app.

host/src/containers/GameSessionContainer.tsx
```diff
@@ -104,20 +112,66 @@ const GameSessionContainer = () => {
...
+          setHeaderGameCurrentTime(response.phaseOneTime);
+        else if (response.currentState === stateArray[6])
+          setHeaderGameCurrentTime(response.phaseTwoTime);
+
         setGameSession(response);
+        checkGameTimer(response);
       });
   };
 
+  const checkGameTimer = (gameSession) =>{
```
maniramezan, line 153: nit: Space after `=>`. There's a way to set the formatting in vscode so it formats for you on save. I don't think we've standardized it and we don't use a common pattern among apps. @lucahendicott might have the `vscode` setting that you can use to format the code for you that matches the pattern we use for host

host/src/containers/GameSessionContainer.tsx
```diff
@@ -104,20 +112,66 @@ const GameSessionContainer = () => {
...
+          setGameTimer(true);
+          setGameTimerZero(false);
+      }
+  };
+
   const handleStartGameModalTimerFinished = () => {
     let newUpdates = {currentState: GameSessionState.CHOOSE_CORRECT_ANSWER, currentQuestionIndex: 0};
     apiClient.updateGameSession({ id: gameSessionId, ...newUpdates })
       .then(response => {
+        localStorage.setItem('currentGameTimeStor', gameSession.phaseOneTime);
```
maniramezan, line 168: typo? `currentGameTimeStore`

host/src/containers/GameSessionContainer.tsx
```diff
@@ -104,20 +112,66 @@ const GameSessionContainer = () => {
...
+      return () => clearInterval(refreshIntervalId);
+    }
+  }, [gameTimer, gameTimerZero, headerGameCurrentTime]);
+
+  useEffect(()=>{ //update gameSession currentTimer every 3 seconds with time from headerGame Timer 
+    if (gameTimer && !gameTimerZero){
+      let refreshIntervalId = setInterval(() => {
+          let newUpdates = {currentTimer: (localStorage.getItem('currentGameTimeStor')>= 0 ? localStorage.getItem('currentGameTimeStor') : 0)};
+          apiClient.updateGameSession({ id: gameSessionId, ...newUpdates });
+      }, 3000);
```
maniramezan, line 134: We should test with different numbers here, especially with mobile devices and the internet access, there can be delays. 3s might just work or be too many updates

maniramezan (changes requested):

networking/src/index.ts
drewjhart: sweet!

host/src/containers/GameSessionContainer.tsx
drewjhart: deleted extra `()`

host/src/containers/GameSessionContainer.tsx
drewjhart: went through `GameSessionContainer`, `StartGame`, `GameInProgress` and applied new line + space after // to longer comments

host/src/containers/GameSessionContainer.tsx
drewjhart: commenting updated and spaces after `=> `added

host/src/containers/GameSessionContainer.tsx
drewjhart: space fixed. those settings would be awesome though

host/src/containers/GameSessionContainer.tsx
drewjhart: good call, variable name changed throughout

host/src/containers/GameSessionContainer.tsx
drewjhart: 100% agree, I think we'll need to build and test with mobile to find the acceptable frequency

host/src/containers/GameSessionContainer.tsx
drewjhart: Ok @maniramezan, I've updated all the comparisons to be like `nextState === GameSessionState.PHASE_1_DISCUSS`. I've removed the `stateArray` and used object.keys to determine the next state from the enum.

networking/src/index.ts
maniramezan, line 4: Can you delete this file locally? If you do and then it won't show up anymore

maniramezan (changes requested):

networking/src/index.ts
drewjhart: shoot, yup that's been done now. PR updated

PR #377, opened 2022-10-25
map wrong answer reasons to individual cards.

lucahendicott:
Map wrong answer reasons to individual cards.
TODO: update carousel to swipe through each card

### DIFF
--- a/mobile/ios/Podfile.lock
+++ b/mobile/ios/Podfile.lock
@@ -381,7 +381,7 @@ PODS:
     - React
   - rn-fetch-blob (0.12.0):
     - React-Core
-  - RNCAsyncStorage (1.17.9):
+  - RNCAsyncStorage (1.17.10):
     - React-Core
   - RNCCheckbox (0.5.12):
     - React-Core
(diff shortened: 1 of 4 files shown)


mobile/ios/Podfile.lock
```diff
@@ -381,7 +381,7 @@ PODS:
     - React
   - rn-fetch-blob (0.12.0):
     - React-Core
-  - RNCAsyncStorage (1.17.9):
```
maniramezan, line 384: Looks like you're updating this dependency without updating the version in `package.json`

mobile/src/screens/JoinGame/index.js
```diff
@@ -48,13 +47,20 @@ export default function JoinGame({
...
                     teamMember,
                 })
 
+            case GameSessionState.PHASE_1_RESULTS:
+            case GameSessionState.PHASE_2_RESULTS:
+                return navigation.navigate("PhaseResult", {
+                    gameSession,
+                    team: gameSession.teams.find((team) => team.id === teamId),
+                    teamMember,
+                })
```
maniramezan, line 62: We'd also need avatars from select team screen to be passed along to all screens

lucahendicott: Got it. Will do.

mobile/ios/Podfile.lock
lucahendicott: Oh ok I think I see it. Will update

PR #398, opened 2022-10-30
Host: first answer issue fix

drewjhart:


host/src/containers/GameSessionContainer.tsx
```diff
@@ -175,7 +179,17 @@ const GameSessionContainer = () => {
         setHeaderGameCurrentTime(gameSession.phaseOneTime);
         checkGameTimer(response);
         setGameSession(response);
-        setIsModalOpen(false); 
+
+        const teamDataRequests = response.teams.map(team => {
+          return apiClient.getTeam(team.id); // got to call the get the teams from the APi so we can see the answers
```
maniramezan, line 184: nit: `i` in API being capital

host/src/containers/GameSessionContainer.tsx
```diff
@@ -49,12 +49,13 @@ const GameSessionContainer = () => {
     let gameSessionSubscription: any | null = null;
     gameSessionSubscription = apiClient.subscribeUpdateGameSession(gameSessionId, response => {
       // only run the gametimer check on instances where the currentState changes (new screens)
-      if (gameSession && gameSession.currentState !== response.currentState) 
+      if (gameSession && gameSession.currentState !== response.currentState){
```
maniramezan, line 52: nit: space before `{`. Same for the below changes in this file

host/src/containers/GameSessionContainer.tsx
```diff
@@ -175,7 +179,17 @@ const GameSessionContainer = () => {
         setHeaderGameCurrentTime(gameSession.phaseOneTime);
         checkGameTimer(response);
         setGameSession(response);
-        setIsModalOpen(false); 
+
+        const teamDataRequests = response.teams.map(team => {
```
maniramezan, line 183: How often are we calling this method? Looks like this method is called when we go to choosing the correct answer. If that's true, then there won't be any answers in the teams

drewjhart: That's correct. everything from 183 - 192 populates a `teamArray` when we go from `StartGame `to `GameInProgress`. The `teamsArray` contains all the `teams`, `teamMembers`, and crucially `teamAnswers` because of that depth issue in the `GameSession`.  

The thinking here is to allow add/drop of teams to occur at the `StartGame` page and then capture a snapshot of the players in the game as we move to `GameInProgress` and put that in the `teamsArray`. The `CreateTeamAnswers` subscription will then operate on the array to track the answers.

There was an issue where the `teamsArray` was blank when we first went to CHOOSE_CORRECT_ANSWER. This is added to ensure the `teamsArray` is populated as we move into the game.

host/src/containers/GameSessionContainer.tsx
drewjhart: fixed

host/src/containers/GameSessionContainer.tsx
drewjhart: fixed

host/src/containers/GameSessionContainer.tsx
maniramezan: Thanks a lot for the details. hmmm, I believe the response to `updateGameSession` doesn't have the depth that we need, but it might already has the teams if not members. The current approach works, but as an alternative, we can also just fetch the whole game session which at this point should include all the teams and update the current game session object to this one. That way we have one source of truth that includes all the info we need.

I don't have any preference here as both approaches have their own pros and cons and at this time, it doesn't matter much for our case.

maniramezan (approved):

host/src/containers/GameSessionContainer.tsx
drewjhart: Definitely a great point - the teamsarray set up is a bit cumbersome as it stands right now.  I've made a note of this in our post-MVP log so we can figure out the best approach following our deadline:

https://docs.google.com/document/d/1icfDBtpmzPHmBHSzMedhNmFElKitVdJGEcFfg4xB12M/edit?usp=share_link

In the meantime, I'll test once more tomorrow and then merge

PR #411, opened 2022-11-02
Mobile: Update phase 2 answer submit API

lucahendicott:
- Set isTrickAnswer to true in phase 2 answer submission. 
- Handle default for more than 5 players in leaderboard.

### DIFF
--- a/mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
+++ b/mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
@@ -39,10 +39,10 @@ const PhaseTwoBasicGamePlay = ({
     const question = gameSession?.isAdvanced
         ? team.question
         : gameSession?.questions[
-        gameSession?.currentQuestionIndex == null
-            ? 0
-            : gameSession?.currentQuestionIndex
-        ]
+              gameSession?.currentQuestionIndex == null
+                  ? 0
+                  : gameSession?.currentQuestionIndex
+          ]
 
     const phaseTime = gameSession?.phaseTwoTime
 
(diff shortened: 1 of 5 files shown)


mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
```diff
@@ -96,7 +96,8 @@ const PhaseTwoBasicGamePlay = ({
                                 teamMember.id,
                                 question.id,
                                 answer.text,
-                                answer.isChosen ? null : false
+                                answer.isChosen ? null : false,
+                                answer.isTrickAnswer ? false : true
```
maniramezan, line 100: I'm a bit confused on these two lines. `answer` here is what we have created locally in this file and it doesn't have `isChosen` or `isTrickAnswer`. Those fields are always `undefined` unless I've missed something here. These two fields should be hardcoded here. Since it's phase 2, the last parameter should always be `true`. `isChosen` is not needed by mobile, unless that's used by the host, it also can be always `false` here

lucahendicott: Hmm ok. `isChosen` is needed by host. It does update the fields in the backend properly for `isChosen` so I just went the same route for `isTrickAnswer` - It's not updating properly in DB but maybe I can follow suit with what Ryan updated in phase 1

maniramezan: `isTrickAnswer` should always be `false` in phase 1 and `true` for phase 2 submissions. If the data is not updated on the backend, I can look into that. Since this field is used for score calculation, setting the wrong value would cause the score to be calculated wrong

PR #440, opened 2022-11-26
[Mobile] Answer choices displaying on Android

drewjhart:


drewjhart:
In response to https://github.com/rightoneducation/righton-app/issues/439

Will need to be tested with iOS. Merging so we can test it at our next meeting

PR #453, opened 2022-12-13
Alex k  mobile image padding

Alexandra-Kushnirsky:


maniramezan:
Lets exclude `Podfile.lock` changes as they shouldn't be here. I believe the `package.json` has some dependencies set up in a way that if you remove `package-lock.json` it updates them. I need to look into it why

PR #458, opened 2022-12-15
ios cannot swipe back now

josephinaim:
I stashed the Podfile.lock changes this time, had to pod update in ios in order for the xcode simulator to run.


drewjhart:
nice!

PR #495, opened 2022-12-22
commented out margin top

josephinaim:
- the game instructions were cropped out because of smaller screen size
- getting rid of the margin top may cause unwanted changes to bigger screen sizes
- podfile.lock still has changes? should i revert and submit a new pr?

### DIFF
--- a/mobile/ios/Podfile.lock
+++ b/mobile/ios/Podfile.lock
@@ -18,6 +18,7 @@ PODS:
   - AppCenterReactNativeShared (4.4.5):
     - AppCenter/Core (= 4.4.3)
   - Base64 (1.1.2)
+  - BEMCheckBox (1.4.1)
   - boost-for-react-native (1.63.0)
   - BVLinearGradient (2.6.2):
     - React-Core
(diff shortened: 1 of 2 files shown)


mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
```diff
@@ -35,6 +35,6 @@ const styles = StyleSheet.create({
         fontFamily: fontFamilies.montserratBold,
         fontSize: fonts.xxMedium,
         color: colors.white,
-        marginTop: 30,
+       // marginTop: 30
```
maniramezan, line 38: Instead of commenting the code, you should always remove it.

mobile/ios/Podfile.lock
```diff
@@ -18,6 +18,7 @@ PODS:
   - AppCenterReactNativeShared (4.4.5):
     - AppCenter/Core (= 4.4.3)
   - Base64 (1.1.2)
+  - BEMCheckBox (1.4.1)
```
maniramezan, line 21: Did you add this new dependency? If not, the changes here shouldn't be needed

maniramezan (changes requested):

PR #499, opened 2022-12-23
Alex  mobile UI enter name

Alexandra-Kushnirsky:


mobile/yarn.lock
```diff
@@ -0,0 +1,11693 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
maniramezan, line 1: Looks like you've used `yarn install` instead of `npm install`

maniramezan (changes requested):

maniramezan (approved):

PR #504, opened 2023-01-03
Alex  mobile enter name app crash

Alexandra-Kushnirsky:


mobile/src/Student/screens/StudentName/index.js
```diff
@@ -12,16 +12,27 @@ const StudentName = ({ navigation, gameSession, setTeamInfo }) => {
...
-        if (!firstName && firstNameTextRef) {
+        /*if (!firstName && firstNameTextRef) {
             firstNameTextRef.focus()
             return
         }
 
         if (!lastName && lastNameTextRef) {
             lastNameTextRef.focus()
             return
+        }*/
```
maniramezan, line 26: Always go with removing the code instead of commenting it. This helps to keep the code clean

mobile/src/Student/screens/StudentName/index.js
```diff
@@ -86,7 +97,8 @@ const StudentName = ({ navigation, gameSession, setTeamInfo }) => {
                                         placeholder={"First Name"}
                                         placeholderTextColor={colors.primary}
                                         ref={(ref) => {
-                                            this.firstNameTextRef = ref
+                                            //this.firstNameTextRef = ref
+                                            firstNameInput = ref
```
maniramezan, line 101: Both here and below, I believe it should still be `firstNameTextRef` as that gives us a references to the text field and then we need to focus on it. Also, where `firstNameInput` is defined? Same is true for the password

maniramezan (changes requested):

drewjhart:
Comments above are great. Also, per our discussion during paired programming, I've tested it in Android and the behaviour looks good. I might be nitpicking here but would it be possible to clear the error message once both of the text fields are populated? It will prevent any confusion if there's a slowdown on the device.  My android simulator is pretty slow but you can check out the video to see there might be some confusion when someone fills out the first/last name and hits submit but still sees the error message.

https://user-images.githubusercontent.com/79417944/210428081-b29b66a5-fdd9-4688-9ee2-1973a71a0566.mp4

mobile/src/Student/screens/StudentName/index.js
Alexandra-Kushnirsky: Yes, will delete before next push! Didn't realize I hadn't deleted it

mobile/src/Student/screens/StudentName/index.js
Alexandra-Kushnirsky: When I was looking into it, it said that 'firstNameTextRef' is declared but its value was never used so I just commented it out. firstNameInput (and lastNameInput) is declared at the top of the file on lines 11 and 12:
"  const [firstName, setFirstName] = useState("")
    const [lastName, setLastName] = useState("") "
I essentially replicated the format that was used for taking in user inputs in the enter game code screen (which used state instead of ref for user input):
"
const [gameCode, setGameCode] = useState("")
const [showErrorText, setShowErrorText] = useState(false)
"

Alexandra-Kushnirsky:
I think getting rid of the error message once both inputs are populated by the user should not be a difficult fix and I will definitely look into it.

drewjhart:
I am closing this PR as it has been superseded by: https://github.com/rightoneducation/righton-app/pull/512

PR #515, opened 2023-01-05
481 mobile UI updates on monster screen s13

josephinaim:
podfile is the same as dev branch


mobile/src/Student/screens/SelectTeam/index.js
```diff
@@ -128,7 +129,21 @@ const styles = StyleSheet.create({
   teamIcon: {
     resizeMode: 'contain',
     flex: 1,
-    alignSelf: 'center'
+    alignSelf: 'center',
+    marginTop: 15,
+    //borderWidth: 7,
```
maniramezan, line 134: nit: Always remove the code instead of commenting it

mobile/src/Student/containers/TeamIcons/img/index.js
```diff
@@ -0,0 +1,158 @@
+import { useState } from 'react'
+import { FlatList, Image, SafeAreaView, StyleSheet, Text, View } from 'react-native'
+import { scale, verticalScale } from 'react-native-size-matters'
+import Pressable from 'react-native/Libraries/Components/Pressable/Pressable'
+import PurpleBackground from '../../../components/PurpleBackground'
+import RoundButton from '../../../components/RoundButton'
+import { colors, fontFamilies, fonts } from '../../../utils/theme'
+import TeamIcons from '../../containers/TeamIcons'
+
+const SelectTeam = ({ navigation, team, saveTeamAvatar }) => {
```
maniramezan, line 10: I think you had this in another PR that we discussed, would you mind reminding me again why we're creating a new class here which doesn't seem it's used anywhere else?

maniramezan (changes requested):

mobile/src/Student/containers/TeamIcons/img/index.js
maniramezan: kk found the response, so if the location changes, then we won't need to create a new file. Or did you just moved the file and I missed the removal of the file? You can use `git mv src dst` to move a file in git. That way it keeps the history of the file as well

maniramezan (changes requested):

mobile/src/Student/screens/SelectTeam/index.js
drewjhart: removed

mobile/src/Student/containers/TeamIcons/img/index.js
drewjhart: So, unless I'm misinterpreting something, I don't think we need this `index.js` specifically under `/TeamsIcons/img`. I think `SelectTeam` happens under `/SelectTeam/index.js` and this might have just been a copy of that` index.js` file moved over inadvertently. I'm basing this off the fact that the relative path on the `TeamIcons` import maps to the correct folder if the `index.js` is located in `/SelectTeam` @josephinaim, is this accurate? I've gone ahead and pushed a commit that adjusts this so if you can confirm that we don't need that `index.js`, I'll merge

josephinaim: Yes, the new commit seems fine on my end!

PR #520, opened 2023-01-09
Alex  mobile answer expl carousel

Alexandra-Kushnirsky:


mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
```diff
@@ -155,20 +195,34 @@ const PhaseTwoBasicGamePlay = ({
...
-            </Card>
-        ))
-    }
+                <ScrollableQuestion question={question} />
+                <View style={styles.roundContainerCorrect}>
+                    <Text style={styles.correctAnswerText}>{correctAnswer.text}</Text>
+                </View>
+                <HintsView hints={availableHints} />
+                <Text style={styles.reasonsText}>{correctAnswer.reason}</Text>
+            </Card></>]
```
maniramezan, line 211: nit: `</>]` should be in a new line

mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
```diff
@@ -181,33 +235,42 @@ const PhaseTwoBasicGamePlay = ({
...
+                            <Text style={styles.cardHeadingText}>Correct Answer</Text>
+                            <ScrollView showsVerticalScrollIndicator={false}>
+                                {correctAnswerScreen}
+                            </ScrollView>
+                        </>
+                        <>
+                            <Text style={styles.cardHeadingText}>Wrong Answers</Text>
+                            <ScrollView showsVerticalScrollIndicator={false}>
+                                {wrongAnswersScreen}
+                            </ScrollView></>
```
maniramezan, line 272: nit: new line for `</>`

mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
```diff
@@ -112,15 +116,51 @@ const PhaseTwoBasicGamePlay = ({
...
     const submittedAnswerText = `Thank you for submitting!\n\nWaiting for your teacher to advance to the next section`
 
-    let cards = [
-        <Card headerTitle="Question">
-            <ScrollableQuestion question={question} />
-        </Card>,
-        <View>
-            <Card headerTitle="Answers">
+    const timerHeading =
+        [<>
```
maniramezan, line 123: I might be wrong, but any reason that this and any other defined view variables here are always inside a `[]` which means they are an array? All views are enclosed inside a fragment so there's always one fragment that encloses the subviews.

maniramezan (changes requested):

mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
Alexandra-Kushnirsky: No, you are correct! I took reference from the cards const that was there previously but this doesn't make sense to use in the case above since nothing is being iterated through. Will remove and fix the code above as well. Thanks!

maniramezan (approved):
Thanks! ❤️

PR #526, opened 2023-01-11
[Mobile] - Change navigation.push to navigation.navigate on gameplay screens

drewjhart:
I'd like to make the included updates because @Alexandra-Kushnirsky and I have been finding that when we use navigation.push, the new instance of the screen resets `selectedAnswerIndex` via `const [selectedAnswerIndex, setSelectedAnswerIndex] = useState(null)` (ln 39 - PhaseTwoBasicGamePlay). 

We're looking to persist that value across the state change from `CHOOSE_TRICKIEST_ANSWER` to `PHASE_2_RESULTS` and the creation of a new instance of the screen seems to prevent that from happening.

I've done a test run through the app with `navigate` and the behaviour between `CHOOSE_CORRECT_ANSWER` and `PHASE_1_RESULTS` and between `CHOOSE_TRICKIEST_ANSWER` and `PHASE_2_RESULTS` seems unaffected.

### DIFF
--- a/mobile/src/screens/JoinGame/index.js
+++ b/mobile/src/screens/JoinGame/index.js
@@ -43,12 +43,12 @@ export default function JoinGame({
 
             case GameSessionState.CHOOSE_TRICKIEST_ANSWER:
             case GameSessionState.PHASE_2_DISCUSS:
-                navigation.push("PhaseTwoBasicGamePlay")
+                navigation.navigate("PhaseTwoBasicGamePlay")
                 break
 
             case GameSessionState.PHASE_1_RESULTS:
             case GameSessionState.PHASE_2_RESULTS:
-                navigation.push("PhaseResult")
+                navigation.navigate("PhaseResult")
                 break
 
             case GameSessionState.FINAL_RESULTS:


mobile/src/screens/JoinGame/index.js
```diff
@@ -43,12 +43,12 @@ export default function JoinGame({
 
             case GameSessionState.CHOOSE_TRICKIEST_ANSWER:
             case GameSessionState.PHASE_2_DISCUSS:
-                navigation.push("PhaseTwoBasicGamePlay")
+                navigation.navigate("PhaseTwoBasicGamePlay")
```
maniramezan, line 46: I highly recommend against this as although phase one and phase two are using the same class instructions, they are two separate screens and should be interacted the same way. If the issue is `selectedAnswerIndex`, then that's something that can be computed for phase 2 based on the data that we have already.

In addition, `navigate` pops back in the stack to that view which messes up the navigation stack and can cause some other issues

mobile/src/screens/JoinGame/index.js
```diff
@@ -43,12 +43,12 @@ export default function JoinGame({
...
                 break
 
             case GameSessionState.PHASE_1_RESULTS:
             case GameSessionState.PHASE_2_RESULTS:
-                navigation.push("PhaseResult")
+                navigation.navigate("PhaseResult")
```
maniramezan, line 51: Same as above

maniramezan (changes requested):

mobile/src/screens/JoinGame/index.js
drewjhart: Good to know, we'll have to find another way. I'm closing this PR without merge. Thanks for clarifying, Mani.

PR #531, opened 2023-01-16
[Mobile] - Multiple Questions & Refactor

drewjhart:
Hi all,
This PR has two goals:

1. **Multiple Questions** - Provide multiple question functionality to `mobile` with basic scoring (Phase 1 correct answer + 10)
2. **Refactor** - Begin refactoring `mobile` for performance

**Multiple Questions**
Code has been adjusted to allow for multiple questions, in navigation, scoring and the resolving of some errors. For the errors, https://stackoverflow.com/questions/52368342/invariant-violation-text-strings-must-be-rendered-within-a-text-component was crucial and you can see the update from `&&` to ternary.

**Refactor**
_Issue:_
We are using `react-navigation` for our stack. A critical point of this package is that screens _remain mounted even when not displayed to the user_ (see: https://reactnavigation.org/docs/navigation-lifecycle/). This means that every time we create a screen, there is the potential for code to be run on that screen even if it is not displayed to the user. 

This surprised me (coming from ReactJs) and means that we need to be pretty careful about rerenders and useEffects in our code, particularly given our program structure. To recap, we currently have: 
image

We subscribe to the entire `gameSession` object in the topmost container and then pass the object down to all the screens every time there is an update to the `gameSession` object. As all the screens stay mounted, that means that every time we are updating the `gameSession` object on the back end, there is the potential for all of our screens to be rerendered. I added `console.logs` in the render method of each of the screens in our existing `dev` branch and you can see how these are getting fired even when they aren't showing to the user. 

https://user-images.githubusercontent.com/79417944/220466852-cd1b45f4-1b40-4692-9f90-829821968fd3.mp4

To be fair, I don't think all of these `console.logs` are equivalent to actual rerenders on the DOM but it's good to note that there's potential for a lot of code to be run in the background here (given the container structure of the app).

_Resolution:_

I am continuing to assess how much _actual_ performance impact there is as a result of this. In the meantime, there are a few areas I've targeted that represent "low-hanging fruit"

1. useEffects/dependency arrays
Based on current comments in the code, it seems like a lot of our useEffects are running in the background even when the screens aren't displayed to the user. `react-navigation` docs suggest using `useFocusEffect` to prevent this (https://reactnavigation.org/docs/use-focus-effect/). I have implemented `useFocusEffect` with `useCallback` to ensure this code is only run when the screen is displayed. I have also revised any dependency arrays to only include the minimum changes and tried to avoid any `[gameSession]`

2. `navigation.navigate` vs `navigate.push`
We sometimes use  `navigate.push` to generate new copies of existing screens (PHASE_RESULTS, for example). While these screens get unmounted when a previous` navigation.navigate` is called (moving form question 1 to question 2, for example), I want to minimize the number of screens in the stack at any given time. I've rewritten our screens to only include `navigate.navigate` and have provided some cleanup functions when screens unfocus on `blur` (as suggested here: https://reactnavigation.org/docs/navigation-events/) to make this work

3. general cleanup
I've tried to clean up any redundancies in code I've come across generally. Examples include hoisting API calls to one central location, rewriting scoring so that each questions score is only calculated once, etc. I don't think these will have huge performance impacts though.

**Next Steps - Multiple Questions:**
Further testing. Exchange basic +10points in Phase 1 with multipliers.

**Next Steps - Refactor:**
This PR was getting large and I'd like to get it in front of other peoples' eyes for feedback and comment before I continue. Following any discussion and rewrites, I'd like to merge and then find a better way to actually gauge performance impacts on the potential rerenders. For example, we currently use a `useEffect` in `JoinGame` for our switch navigation and there's got to be a better way to locate that centrally. 

I've already tested the code with https://www.npmjs.com/package/@welldone-software/why-did-you-render and it didn't generate anything, so I'll need to continue to dig. Everyone's thoughts more than appreciated on this one!


mobile/src/Student/screens/EnterGameCode/index.js
```diff
@@ -69,11 +69,11 @@ const EnterGameCode = ({ navigation, fetchGameSessionByCode, handleSubscribeToGa
                             style={styles.enterButton}
                             onPress={this.onGameCodeSubmit}
                         />
-                        {showErrorText && <Text
+                        {showErrorText ? <Text
```
maniramezan, line 72: Any reason for replacing this with `?` and the previous one with `&&`?

drewjhart: Thanks for taking a look at this. I've replaced ` &&` with the ternary because I was getting an "Invariant Violation: Text strings must be rendered within a <Text> component" error when using games with multiple questions. 

After some research, I found this post on Stack Overflow: https://stackoverflow.com/questions/52368342/invariant-violation-text-strings-must-be-rendered-within-a-text-component

Replacing each of the conditional renders with the ternary resolved the issue with the Invariant Violation. Let me know if you have any feedback on this though.

host/src/pages/GameInProgress.jsx
```diff
@@ -107,7 +107,7 @@ export default function GameInProgress({
           team.teamMembers && team.teamMembers.forEach(teamMember => {
             teamMember.answers && teamMember.answers.forEach(answer =>{
             if (answer.questionId === currentQuestionId){
-               if (((currentState === GameSessionState.CHOOSE_CORRECT_ANSWER || currentState === GameSessionState.PHASE_1_DISCUSS) && answer.isChosen) || ((currentState === GameSessionState.PHASE_2_DISCUSS || currentState === GameSessionState.CHOOSE_TRICKIEST_ANSWER) && !answer.isChosen)){
+               if (((currentState === GameSessionState.CHOOSE_CORRECT_ANSWER || currentState === GameSessionState.PHASE_1_DISCUSS) && answer.isChosen) || ((currentState === GameSessionState.PHASE_2_DISCUSS || currentState === GameSessionState.CHOOSE_TRICKIEST_ANSWER) && answer.isTrickAnswer)){
```
drewjhart, line 110: identify incoming answers to `host` in Phase 2 using `answer.isTrickAnswer`

mobile/src/Navigator/index.js
```diff
@@ -28,6 +28,9 @@ const AppContainer = () => {
                 teamAvatar,
                 saveTeamAvatar,
                 clearStorage,
+                handleSubscribeToGame,
+                handleAddTeam,
+                handleAddTeamAnswer
```
drewjhart, line 33: hoisted these functions from individual components to container so we don't need to duplicate API code. There might be a rerender cost to this but I feel like it's cleaner

mobile/src/Student/components/Answers/index.js
```diff
@@ -54,7 +54,7 @@ const Answers = ({ teamSelectedTrickAnswer }) => {
     <View
       style={[styles.mainContainer, { height: scale(65 * answers.length) }]}
     >
-      {answers &&
+      {answers ?
         answers.map((ans, index) => {
```
drewjhart, line 58: see discussion below re: ternaries and multiple questions

mobile/src/Student/components/LoadingIndicator/index.js
```diff
@@ -1,4 +1,5 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState } from 'react'
+import { useFocusEffect } from '@react-navigation/native'
 import { View, StyleSheet, Text } from 'react-native'
```
drewjhart, line 3: using useFocusEffect to prevent useEffects running on screens that aren't focused

mobile/src/Student/components/LoadingIndicator/index.js
```diff
@@ -107,7 +93,9 @@ export default LoadingIndicator = (
         return () => {
             clearInterval(refreshIntervalId)
         }
-    })
+      },[remainingSecondsInMilliSeconds])
+    )
+
     return (
```
drewjhart, line 99: trying to limit the running of the countdown as much as possible

mobile/src/Student/containers/GameSessionContainer.js
```diff
@@ -76,71 +77,6 @@ const GameSessionContainer = ({ children }) => {
     const [teamMember, setTeamMember] = useState(null)
     const [teamAvatar, setTeamAvatar] = useState(TeamIcons[0])
 
-    useEffect(() => {
```
drewjhart, line 79: removing local storage code for now

mobile/src/Student/containers/GameSessionContainer.js
```diff
@@ -160,23 +96,85 @@ const GameSessionContainer = ({ children }) => {
...
             global.apiClient.subscribeUpdateGameSession(
                 gameSession.id,
                 (gameSessionResponse) => {
                     setGameSession(gameSessionResponse)
+                    console.log("subscription triggered")
+                    console.log(gameSessionResponse.currentState)
                 })
+    }
+
+    const handleAddTeam = async (teamName) =>
```
drewjhart, line 115: hoisted api calls

mobile/src/Student/containers/GameSessionContainer.js
```diff
@@ -160,23 +96,82 @@ const GameSessionContainer = ({ children }) => {
             })
     }
 
-    useEffect(() => {
```
drewjhart, line 163: removing the useEffect so the subscription only happens once

mobile/src/Student/screens/Game/Components/HintsView/index.js
```diff
@@ -12,7 +12,7 @@ const HintsView = ({ hints }) => {
     })
 
     return mappedHints.map((item) => (
-        <Hint hintNo={item.hintNo} hint={item.hint} />
+        <Hint hintNo={item.hintNo} hint={item.hint} key={item.hintNo}/>
```
drewjhart, line 15: keys added to removing warnings about unique key requirements

mobile/src/Student/screens/PhaseResult/index.js
```diff
@@ -1,69 +1,62 @@
...
-        setSelectedTrickAnswer(ModelHelper.getSelectedTrickAnswer(updatedCurTeam, curQuestion.id))
-        setCorrectAnswer(ModelHelper.getCorrectAnswer(curQuestion))
-        setCurrentQuestion(curQuestion)
+
+        setSelectedTrickAnswer(ModelHelper.getSelectedTrickAnswer(updatedCurTeam, currentQuestion.id))
         setLoadedData(true)
         setTeamInfo(updatedCurTeam, updatedCurTeam.teamMembers[0])
-    }, [gameSession])
+      }, [gameSession?.currentState])
+    )
```
drewjhart, line 56: this is a fairly heavy useEffect so I only want to update this on currentState changes, and only when the screen is focused. 
 TODO: I think we can continue to optimize this because a lot of the existing calls on the screen are being duplicated in ModelHelper.

networking/src/ModelHelper.ts
```diff
@@ -100,20 +100,18 @@ export abstract class ModelHelper {
...
+        if (answers.some(answer => answer?.isTrickAnswer===true)){
+          let newScore = originalScore += ModelHelper.calculateBasicModeWrongAnswerScore(gameSession, answers?.find(answer => answer?.isTrickAnswer==true)?.text ?? '', currentQuestion.id)
+          return newScore
+        }
+        else if (answers.some(answer => answer?.isChosen===true && answer?.text == correctAnswer?.text)){
+          let newScore = originalScore += this.correctAnswerScore
+          return newScore
+        }
+        return originalScore
     }
```
drewjhart, line 115: I rewrote this to use team.score to calculate the ongoing score. I think this might be smoother than having to recalculate the score of all questions using a .reduce

networking/src/ModelHelper.ts
```diff
@@ -100,20 +100,18 @@ export abstract class ModelHelper {
...
-            if (answer.isTrickAnswer) {
-                return score + this.calculateBasicModeWrongAnswerScore(gameSession, answer.text, question.id)
-            } else {
-                console.log(`${answer.text} === ${correctAnswer?.text} = ${answer.text === correctAnswer?.text}`)
-                return score + (
-                    answer.text === correctAnswer?.text ? this.correctAnswerScore : 0)
-            }
-        }, 0)
+        if (answers.some(answer => answer?.isTrickAnswer===true)){
+          let newScore = originalScore += ModelHelper.calculateBasicModeWrongAnswerScore(gameSession, answers?.find(answer => answer?.isTrickAnswer==true)?.text ?? '', currentQuestion.id)
```
maniramezan, line 107: No need for `==true`

networking/src/ModelHelper.ts
```diff
@@ -100,20 +100,18 @@ export abstract class ModelHelper {
...
+          return newScore
+        }
+        else if (answers.some(answer => answer?.isChosen===true && answer?.text == correctAnswer?.text)){
+          let newScore = originalScore += this.correctAnswerScore
+          return newScore
```
maniramezan, line 112: You can also just return it instead of putting it into a variable

networking/src/ModelHelper.ts
maniramezan, line 107: Probably cleaner and more readable if you create a new variable that's assigned to the result of `ModelHelper.calculateBasicModeWrongAnswerScore(gameSession, answers?.find(answer => answer?.isTrickAnswer==true)?.text ?? '', currentQuestion.id)` and then just return `originalScore + newVar`

networking/src/ModelHelper.ts
```diff
@@ -100,20 +100,18 @@ export abstract class ModelHelper {
         }
 
         const correctAnswer = this.getCorrectAnswer(question)
+        const currentQuestion = gameSession?.questions[gameSession?.currentQuestionIndex ?? 0]
+        let originalScore = gameSession?.teams?.find(teamElement => teamElement.id === team.id)?.score ?? 0
```
maniramezan, line 104: Why not just use the `team` object that's passed in?

networking/src/ModelHelper.ts
maniramezan: Can you also write some tests for these changes and also make sure the existing tests pass?

mobile/src/components/TeamFooter/index.js
```diff
@@ -1,8 +1,10 @@
-import { Image, StyleSheet, Text, View } from 'react-native'
+import { Animated,  Image, StyleSheet, Text, View } from 'react-native'
+import {useRef, useEffect} from 'react'
 import { scale, verticalScale } from 'react-native-size-matters'
 import { colors, fontFamilies, fonts, fontWeights } from '../../utils/theme'
 
-const TeamFooter = ({ icon, name, totalScore }) => {
+const TeamFooter = ({ icon, name, totalScore, isAnswerCorrect }) => {
```
maniramezan, line 6: Is `isAnswerCorrect` used anywhere in the file? Looks like it's added but not used

mobile/src/components/RoundButton/index.js
```diff
@@ -5,7 +5,7 @@ import { colors, fontFamilies, fonts } from '../../utils/theme'
 
 RoundButton.propTypes = {
     title: PropTypes.string.isRequired,
-    onPress: PropTypes.func.isRequired,
```
maniramezan, line 8: Was there any place in the code that used this component without passing in `onPress`?

mobile/src/Student/screens/SelectTeam/index.js
```diff
@@ -66,7 +65,7 @@ const SelectTeam = ({ navigation, team, saveTeamAvatar }) => {
         <View style={styles.teamIconNameContainer}>
           <Image style={styles.largeIcon} source={avatar && avatar.largeSrc} />
           <Text style={styles.fullNameText}>
-            {team.name}
+            {team ? team.name: null}  {/*propertyuuid doesn't exist*/}
```
maniramezan, line 68: I might be missing something here, but the comment is about uuid while this is using the name

mobile/src/Student/screens/Game/Leaderboard/index.js
```diff
@@ -21,11 +21,7 @@ const Leaderboard = ({
     const sortedTeamsByScore = gameSession.teams.sort((a, b) => b.score - a.score)
 
     const teamName = team.name ? team.name : "Team Name"
-    const totalScore = ModelHelper.calculateBasicModeTotalScoreForQuestion(
-        gameSession,
-        gameSession.questions[gameSession.currentQuestionIndex],
-        team
-    )
+    const totalScore = team.score ? team.score : 0
```
maniramezan, line 24: `score` is not nullable. Is this for checking `team` to be `null`?

mobile/src/Student/containers/GameSessionContainer.js
```diff
@@ -160,23 +96,82 @@ const GameSessionContainer = ({ children }) => {
...
-        if (isNullOrUndefined(gameSession)) {
-            clearStorage()
+    const handleSubscribeToGame = (gameSession) => {
+          if (isNullOrUndefined(gameSession)) {
```
maniramezan, line 100: nit: extra space before `if`

mobile/src/Student/containers/GameSessionContainer.js
```diff
@@ -160,23 +96,82 @@ const GameSessionContainer = ({ children }) => {
...
+                    .addTeamMemberToTeam(team.id, true, uuid.v4())
+                    .then((teamMember) => {
+                        if (!teamMember) {
+                            console.error("Failed to add team member")
+                            return
+                        }
+
+                        if (isNullOrUndefined(team.teamMembers)) {
+                            team.teamMembers = [teamMember]
+                        }
```
maniramezan, line 135: We're missing the `else` condition here to add the member to `team.teamMembers`

mobile/src/Student/containers/GameSessionContainer.js
```diff
@@ -160,23 +96,82 @@ const GameSessionContainer = ({ children }) => {
...
+
+    const handleAddTeamAnswer = async (question, answer, gameSessionState) =>
+    {
+      return  (global.apiClient
+                  .addTeamAnswer(
+                      teamMember.id,
+                      question.id,
+                      answer.text,
+                      (gameSessionState === GameSessionState.CHOOSE_CORRECT_ANSWER && true || gameSessionState === GameSessionState.CHOOSE_TRICKIEST_ANSWER && false),
+                      (gameSessionState === GameSessionState.CHOOSE_CORRECT_ANSWER && false || gameSessionState === GameSessionState.CHOOSE_TRICKIEST_ANSWER && true)
```
maniramezan, line 155: A bit confused here, what's the purpose of `&& true` and `&& false`

maniramezan (changes requested):

mobile/src/Student/containers/GameSessionContainer.js
drewjhart: resolved

mobile/src/Student/screens/Game/PhaseOneBasicGamePlay/index.js
```diff
@@ -12,6 +11,7 @@ import LinearGradient from "react-native-linear-gradient"
 import * as Progress from "react-native-progress"
 import { scale, verticalScale } from "react-native-size-matters"
 import uuid from "react-native-uuid"
+import {useFocusEffect } from '@react-navigation/native'
```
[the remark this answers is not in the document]
drewjhart: resolved

mobile/src/components/TeamFooter/index.js
drewjhart: Not currently, but I think we'll need it when we add in the animation for a correct answer (which is on the list). I'm removing it for now for clarity and will add it back when it's actually required.

networking/src/ModelHelper.ts
drewjhart: resolved, using team.score in place of originalScore

networking/src/ModelHelper.ts
```diff
@@ -100,20 +100,18 @@ export abstract class ModelHelper {
...
-
-            if (answer.isTrickAnswer) {
-                return score + this.calculateBasicModeWrongAnswerScore(gameSession, answer.text, question.id)
-            } else {
-                console.log(`${answer.text} === ${correctAnswer?.text} = ${answer.text === correctAnswer?.text}`)
-                return score + (
-                    answer.text === correctAnswer?.text ? this.correctAnswerScore : 0)
-            }
-        }, 0)
+        if (answers.some(answer => answer?.isTrickAnswer===true)){
```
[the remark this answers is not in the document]
drewjhart: resolved throughout section of code changes

networking/src/ModelHelper.ts
drewjhart: resolved. removed ===true conditions from `anwer?.some()` above as well

networking/src/ModelHelper.ts
drewjhart: resolved

mobile/src/Student/containers/GameSessionContainer.js
```diff
@@ -160,23 +96,82 @@ const GameSessionContainer = ({ children }) => {
...
+                    })
+            })
+            .catch((error) => {
+                console.error(error)
+            })
+    }
+
+    const handleAddTeamAnswer = async (question, answer, gameSessionState) =>
+    {
+      return  (global.apiClient
```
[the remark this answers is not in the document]
drewjhart: resolved

mobile/src/Student/containers/GameSessionContainer.js
drewjhart: Good point. Rewritten for clarity using ternary. the goal here is to ensure that on `CHOOSE_CORRECT_ANSWER`, `isChosen = true` and `isTrickAnswer = false,` while on `CHOOSE_TRICKIEST_ANSWER`, `isChosen=false` and `isTrickAnswer = true`

networking/src/ModelHelper.ts
drewjhart: This was a great comment. Looking into the tests exposed some errors in the logic that I hadn't caught. I've updated the tests to reflect single question and multiple question cases and they both pass.

PR #532, opened 2023-01-16
Alex  mobile UI how to play

Alexandra-Kushnirsky:
(tested on Iphone13 mini and Iphone14, image placement and phone center look consistent)


mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
```diff
@@ -24,16 +30,54 @@ const styles = StyleSheet.create({
         alignContent: 'center',
         backgroundColor: 'rgba(0,0,0,0)',
     },
-    image: {
-        // width: win.width,
-        // height: win.width,
+    phone: {
+        //width: win.width,
+        //height: win.width,
```
maniramezan, line 35: nit: No need to keep the comments, lets remove them

mobile/src/Student/screens/StudentGameIntro/index.js
```diff
@@ -122,7 +137,7 @@ const styles = StyleSheet.create({
     headerTeam: {
         color: colors.white,
         fontFamily: fontFamilies.montserratBold,
-        fontSize: fonts.xxLarge,
+        fontSize: 42,
```
maniramezan, line 140: Lets add this to `theme.js` and use `fonts.` to fetch it. This helps us to have all font sizes in one place and add dynamic text sizing for accessibility later.

mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
```diff
@@ -24,16 +30,54 @@ const styles = StyleSheet.create({
...
+    },
+    imgContainer: {
+        flexDirection: 'row'
+    },
     text: {
         textAlign: 'center',
         textAlignVertical: 'bottom',
         fontFamily: fontFamilies.montserratBold,
-        fontSize: fonts.xxMedium,
+        fontSize: 20,
```
maniramezan, line 50: Lets add this to `theme.js` and use `fonts.` to fetch it. This helps us to have all font sizes in one place and add dynamic text sizing for accessibility later. You can also use the closes one in there instead of adding a new one

mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
```diff
@@ -1,12 +1,18 @@
...
 const win = Dimensions.get('window');
 
-const IntroInfo = ({ image, text }) => {
+const IntroInfo = ({ screen, Monsters, phone, sticker, text }) => {
     return (
         <View style={styles.container}>
-            <Image source={image} style={styles.image} />
+            <Image source={Monsters[0]} style={screen === 1 ? styles.monster1 : styles.monster3} />
+            <View style={styles.imgContainer}>
+                {text != "The most total points wins!" && <Image source={sticker} style={styles.phaseSticker} />}
```
maniramezan, line 12: You shouldn't compare the string in this way. Also, it's error-prune as changing the text in the other screen breaks this. Instead, pass in an extra boolean parameter and use that. That'd be also more readable

mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
```diff
@@ -1,12 +1,18 @@
...
+import { scale, verticalScale } from "react-native-size-matters"
 import { fontFamilies, fonts, colors } from '../../../../utils/theme'
 const win = Dimensions.get('window');
 
-const IntroInfo = ({ image, text }) => {
+const IntroInfo = ({ screen, Monsters, phone, sticker, text }) => {
     return (
         <View style={styles.container}>
-            <Image source={image} style={styles.image} />
+            <Image source={Monsters[0]} style={screen === 1 ? styles.monster1 : styles.monster3} />
```
maniramezan, line 10: using screen to handle the styles to be used here kinda make this component to be tightly coupled with its parent and not reusable for any other cases. Probably a better design can be passing the styles instead of `screen` property so this component can just use them.

maniramezan (changes requested):

mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
```diff
@@ -1,18 +1,18 @@
...
-import { fontFamilies, fonts, colors } from '../../../../utils/theme'
+import { fontFamilies, fonts, colors, fontWeights } from '../../../../utils/theme'
 const win = Dimensions.get('window');
 
-const IntroInfo = ({ screen, Monsters, phone, sticker, text }) => {
+const IntroInfo = ({ monsterStyles, Monsters, phone, sticker, text, showSticker }) => {
     return (
         <View style={styles.container}>
-            <Image source={Monsters[0]} style={screen === 1 ? styles.monster1 : styles.monster3} />
+            <Image source={Monsters[0]} style={monsterStyles[0]} />
```
maniramezan, line 10: fyi: There's this concept called defensive coding which means you try to guard your code always against any failures and crashes, sometimes even going stream and guarding it against something that in theory should never happens.

This applies here to `Monsters` variable. Several notes on it:

- `Monsters` variable can be similar to others and follows camel-based notation, `monsters`
- Check `Monsters` length to make sure there are at least two items in it before accessing them. This can be done as part of adding assertions before `return (`

maniramezan (approved):

mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
Alexandra-Kushnirsky: Thanks, will definitely keep this in mind! Would it be worthwhile to add a try-catch statement, change to camel-case and repush?

maniramezan: I'm ok with the current state, your choice if you'd like to do the changes and push or go with the existing one

mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
```diff
@@ -4,15 +4,18 @@ import { scale, verticalScale } from "react-native-size-matters"
 import { fontFamilies, fonts, colors, fontWeights } from '../../../../utils/theme'
 const win = Dimensions.get('window');
 
-const IntroInfo = ({ monsterStyles, Monsters, phone, sticker, text, showSticker }) => {
+const IntroInfo = ({ monsterStyles, monsters, phone, sticker, text, showSticker }) => {
+    if ({ monsterStyles }.length != 2) {
+        console.debug("'monsterStyles' must be an array of 2 entries")
```
maniramezan, line 9: Since this can cause the app crashes, for future, probably better to use `console.error` and also return an empty `View` otherwise this still causes the app crashes in stores.

nit: You can omit `{}` around `monsterStyles` and I believe you meant checking both `monsterStyles` and `monsters`

maniramezan (approved):

mobile/src/Student/screens/StudentGameIntro/IntroInfo/index.js
Alexandra-Kushnirsky: Thank you!

maniramezan (approved):

PR #535, opened 2023-01-17
Josephina  mobile item 485

josephinaim:
Stars will not be implemented this time around, but spacing and card size issues fixed

### DIFF
--- a/mobile/src/Student/screens/Game/PhaseOneBasicGamePlay/index.js
+++ b/mobile/src/Student/screens/Game/PhaseOneBasicGamePlay/index.js
@@ -331,7 +331,7 @@ const styles = StyleSheet.create({
     hintsViewCard: {
         marginTop: -verticalScale(40),
         paddingVertical: 0,
-        maxHeight: verticalScale(400)
+        maxHeight: verticalScale(500)
     },
     hintsViewQuestion: {
         paddingVertical: verticalScale(30)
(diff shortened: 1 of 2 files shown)


mobile/src/Student/screens/Game/PhaseOneBasicGamePlay/index.js
```diff
@@ -353,7 +353,7 @@ const styles = StyleSheet.create({
     hintsViewCard: {
         marginTop: -verticalScale(40),
         paddingVertical: 0,
-        maxHeight: verticalScale(400)
```
maniramezan, line 356: Should the maximum height to be relative to the screen size? Wondering how this would look on different phone sizes, especially Android that has a lot of variety. Also, we're supporting iPad and Chromebook which wouldn't cause this to be adaptive to those screens

josephinaim: Oh, should I keep this in?

maniramezan: I think for now, we can go with your changes and merge it in, this probably need more broader changes in other screens as well. Just wanted to let you know about things that we should consider when changing sizes

maniramezan (approved):

mobile/src/Student/screens/Game/PhaseOneBasicGamePlay/index.js
josephinaim: Okay, sounds good. It's good to keep in mind as we move into the responsive design for larger screens for sure, thank you!

PR #540, opened 2023-01-20
Alex  mobile wrong ans carousel UI

Alexandra-Kushnirsky:
Was having some trouble implementing automatic scrolling to selected wrong answer as calling "scrollTo(..)" usually requires a button action and this is not what we want in this case. However, the blue circle does appear next to the selected wrong answer choice on the card as in the figma so I thought it would be best to timebox the additional scrolling feature and at least get the blue circle in the updated ui.


maniramezan:
hmm, what's the `scrollTo(..)` issue that you mentioned above? Can you explain in more details?

Alexandra-Kushnirsky:
Sure! So essentially what we need is for the right page in the carousel to automatically scroll to the card corresponding to the student's selected answer choice for what they think is the most popular wrong answer. I wasn't sure how to implement this since the examples I've found of the scrollTo function are triggered by a button (onSubmit) but for the case above, we don't need a button as it should just be automatic. Here is the specific screen I am referring to:
Screenshot 2023-01-23 at 12 29 25 PM

maniramezan:
Thanks a lot for the details! So when should the scroll to happen? Is there an event that happens before the need to scroll? 

The reason you're seeing most samples on `onSubmit` is because probably that's easier to show in a demo app and that's mostly the use-case for scroll to the next section when a button is clicked, but there's nothing that stops you from calling that method at anytime in your view lifecycle

Alexandra-Kushnirsky:
`<ScrollView 
                                showsVerticalScrollIndicator={false} 
                                {...scrollTo({x: 0, y: 0, animated: true})}>
                                {wrongAnswersScreen}
                            </ScrollView>`
Hm I see what you're saying, going to try the above and see how that affects the scrollview

Alexandra-Kushnirsky:
In response to your questions: the page should just automatically scroll to the card in question when the carousel appears (upon entering phase_2_discuss). I don't think there is an event on the mobile side that would trigger this as host is responsible for state changes

maniramezan:
`useEffect` is one place to do if you want to run a code when the component is mounted to the screen. Have you tried that method?

drewjhart:
I am closing this PR as I have integrated the changes into https://github.com/rightoneducation/righton-app/pull/531

PR #541, opened 2023-01-20
[Central] - game/question creation

drewjhart:


drewjhart:
Client-side resolution of create games/questions issue. Replaces any `'` with unicode variant and ensures that `choices` array includes `isAnswer` values for all entries (these were the two main issues with game/question creation.

central/src/App.tsx
```diff
@@ -99,7 +99,7 @@ function App() {
       domain: game.domain,
       cluster: game.cluster,
       standard: game.standard,
-      questions: questionIDSet,
+      questions: questions,
```
drewjhart, line 102: the `updateGame` mutation requires question objects instead of the `questionIDSet` value that we're using in other places in `central`

PR #559, opened 2023-03-10
[Mobile] - Android crashes and card scrolling

drewjhart:
Prior to our demo, we noticed some crashes and scrolling issues on the Android side of the `mobile` app.

There is a crash associated with `TextInputs` in a `FlatLiast` which impacts `PhaseOneGamePlay` and `PhaseTwoGamePlay`. More information can be found here: 
https://github.com/facebook/react-native/issues/17530

This involves the styling of underline properties in a `TextInput` producing `NullPointerException` and hasn't been clearly addressed in the above issue to RN. I implemented the suggested solution (specifying the styling in the `xml`) and still ran into the issue. We don't actually need `TextInput` here though, so I've substituted them for `Text` and that resolved the issue. 

There were also some issues with scrolling not working on Android. This PR also resolves that.

### DIFF
--- a/mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
+++ b/mobile/src/Student/screens/Game/PhaseTwoBasicGamePlay/index.js
@@ -158,8 +158,7 @@ const PhaseTwoBasicGamePlay = ({
                   style={{alignContent: 'center', marginBottom: verticalScale(50)}}
                   showsVerticalScrollIndicator={false}
               >
-                
-                    <Question question={question} />
+                <Question question={question} />
               </ScrollView>
             </Card>
         </>


mobile/android/app/build.gradle
```diff
@@ -138,7 +138,7 @@ android {
         applicationId "com.rightonnew"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 6
+        versionCode 8
```
drewjhart, line 141: This is the correct build number (and version 8 has been built through Android Studio). There was some last minute builds trying to get better crash reporting through bugsnap to understand the crash more but ultimately it was simpler to just swap out the unneeded and problematic `TextInput`

PR #519, opened 2023-01-07
[Mobile] Update RN to 0.70.6 And Add CI/CD

maniramezan:
* Recreate native projects with 0.70.6
* Move to `yarn` instead of `npm`
* Some code cleanup

PR #570, opened 2023-03-21
[Play] - UI for Gameplay Screen

drewjhart:
Replaced by:  https://github.com/rightoneducation/righton-app/pull/575, https://github.com/rightoneducation/righton-app/pull/580, https://github.com/rightoneducation/righton-app/pull/585.

Closing this but pertinent discussion below regarding `yarn` and `TypeScript`


play/.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
maniramezan, line 1: You can use https://www.toptal.com/developers/gitignore/ to generate this file that includes more files and supports Windows hidden files as we ll

play/.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
maniramezan, line 1: Are we using `npm` for this new one? If yes, probably makes sense to switch to yarn

play/build-storybook.log
```diff
@@ -0,0 +1,364 @@
+info @storybook/react v6.5.16
```
maniramezan, line 1: Probably won't need this file

play/package-lock.json
maniramezan, line 1: If you decided to use yarn, you can remove this file

play/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "Right On",
```
maniramezan, line 2: Do we need `!` at the end?

play/src/App.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
```
maniramezan, line 1: If we're setting up a new app, might worth exploring using Typescript instead of jsx

play/public/manifest.json
drewjhart: Good point. This follows `central` and `host` but I can update the three of them to include it.

play/.gitignore
drewjhart: I've updated the file using gitignore.io (specifically `gi macos,windows,react >> .gitignore`) and double checked it against the `.gitignore` files for `host` and `central`. I don't see a specific support for windows hidden files in .gitignore.io but I've added in a couple lines to the bottom that follows: https://stackoverflow.com/questions/8021441/how-to-ignore-all-hidden-directories-files-recursively-in-a-git-repository

let me know if this is too much a blunt instrument for the task though

play/.npmrc
drewjhart: I can switch over to `yarn` for this app. I think that means we'll need to switch over the `yarn` for `central` and `host` as well though. I know we were planning to do that but I'll have to make a point of it prior to any coops joining so don't have confusion. This is more of a self reminder for me, as I'm going to continue focusing on play for now.

maniramezan: I have this work done for mobile, networking, and debug panel. Usually the steps are:

* Remove `package-lock.json`
* Run `yarn install`
* Merge the new changes including `yarn.lock`

For Host app, it'd need the changes to `package.json` script section. You can check [this section](https://github.com/rightoneducation/righton-app/pull/519/files#diff-22d39eaa038d1dbabef4c120dd46c830896fdada76c56b7a7757c1ed5e5bb1c5R28-R29) in my PR.

play/src/App.jsx
maniramezan: Great points and thanks for going in details! I totally agree with this slowing us down and can be harder for others to ramp and start contributing to it. However, this also can be seen as an investment, especially that we're considering PWA version of the Play which means this product would become both our website and mobile version of this product. Especially with the benefits of nullability checks and being strongly types, this can help reducing bugs over time.

We should look at this as an investment to help in future to reduce bugs, take advantage of TS features to have cleaner code, and something that attracts more talent for this product as it offers modern technologies to learn and work on.

With all that said, since you're the one working on it, you should choose the technology that you feel comfortable with.

maniramezan:
Also just a thought, you might wanna start merging this PR when the base is ready and then continue working on new features and screens as follow up PRs. That helps to 
- Have PR reference for each work
- Easier to review the changes
- Easier to refer back to each changes in future to know where the changes are coming from and read the comments

play/.npmrc
drewjhart: Awesome, I'll check that out. Thanks Mani

play/src/App.jsx
drewjhart: That's a great way to frame the issue. I'm going to run this by Sinclair tomorrow and if he signs off begin transferring to TS. Thanks for the insight on this - super valuable.

drewjhart:
> Also just a thought, you might wanna start merging this PR when the base is ready and then continue working on new features and screens as follow up PRs. That helps to
> 
> * Have PR reference for each work
> * Easier to review the changes
> * Easier to refer back to each changes in future to know where the changes are coming from and read the comments

It's a good call for sure. My thinking here is to focus this PR on the UI elements for the first gameplay screen (I'll update the title when it's out of draft so that it's not so vague) so it will include header and footer components, cards, the timer and buttons. Basically these elements:

image

Hopefully this will be enough for U/X coordination but without overwhelming anyone on reviews. Totally let me know what you think though. I don't want to issue some huge, janky PR, at which point you take one look and then close your laptop forever hahaha

maniramezan:
aw great, yeah generally, it'd be helpful if the PR / issue mentioned in the PR includes:
- Screenshots of the screen it's trying to implement
- Broken into small pieces, like this PR includes both project set up and screen work, but these two could be broken into two separate PRs
- Some screens are simple, but some like this are also complicated, this can further down be broken down into header / body / footer PRs
- For yourself not being blocked by these, one clever way is create a new branch from your current branch and then open the PR and choose your last PR as the main branch. This way, it only shows the changes between the two PRs. For exmaple, you can create the infra for the app and open a PR. And then continue the work by creating a new branch and add the code for play screen. That way you're not blocked by code not being merged, but also breaking the implementation into smaller pieces.

Let me know if you wanted to talk about these and different approaches and we can meet and dive deep

play/src/App.jsx
sincwu: Great discussion -- thanks to you both for sharing these details. Spoke with Drew earlier, and given all the pros summarized above, TypeScript sounds like the way go -- full speed ahead!

PR #585, opened 2023-04-12
[Play] - GameInProgress Screen

drewjhart:
**GameInProgress Screen - UI**
Adds UI elements/layout for GameInProgress screen.

**Updates:**
- Page layout has been produced, largely using `Grid`, `Stack` and `Box` components from `@mui`. 
- `Swiperjs` has been added for the carousel on the mobile side 
- Mobile vs Desktop is detected using `useMediaQuery` in `@mui`
- The common styled `paper` components have been moved to `/lib/styledcomponents` for portability
- Theme updates have continued, centralising colors/fonts. Usage examples for each color have been added as comments (the list is getting a bit long)
- Readme.md file has been updated for clarity

This has been built and uploaded to: [play.rightoneducation.com](http://play.rightoneducation.com/)

Storybook can be found here: [Storybook](https://64220218d457f84ba4546eb1-yqqguzhiiq.chromatic.com/?path=/story/design-system-1-atoms-scoreindicator--default)

Tested on: 
- Chrome 
- Safari
- iPhone S
- iPhone 12
-  Android Galaxy A11

You can find some reference material below:
1. Figma Design - screenshots and link to Figma design screen this PR targets
2. Layout Diagrams - diagrams break down the screen section by section, in terms of the code elements used.
3. Samples - a video and screenshot 

**Figma Design:**
image
https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?node-id=13504%3A138577&t=yvbtWLBV7lorFWwV-1

**Layout Diagrams:**
Main container layout:
`Stack` with `Stack item` for header, body and footer
image

Body:
The `body` is another `Stack` with containing the `BodyBoxUpper`, `BodyBoxLower`, and `BodyGridArea`. `BodyBoxUpper` is the purple gradient with box shadow, `BodyBoxLower` is the white background to the footer, and `BodyGridArea` holds the cards and titles with `position: absolute` to float over the background elements. 
image

Body Grid Area:
This layout changes whether desktop or mobile based on `useMediaQuery` from `mui`. Here's a breakdown of each:
Desktop:
`Grid` with each `Grid item` set at `6`:
image

Mobile:
`Grid` with first `Grid item` set at `12` and second at `0`. `swiperjs` swiper is conditionally rendered in first `Grid item` with settings per below:
image

**Samples:** 
Video:
https://user-images.githubusercontent.com/79417944/231609678-d28d9249-136f-4764-bdb0-53c899b6cde2.mp4

Desktop:
image

Mobile:
image


drewjhart:
U/X Coordination: 
Hi @amarax, one item I wanted to get out in front of. You'll notice that the pagination bullets are circles instead of squares. Do you think this is ok? `Swiperjs` I think is our best option for a carousel but the ability to customize the pagination bullets requires more effort. If you think it's important, I can try and match but didn't want to invest the time digging through the docs and developing a custom component if you thought the circles were ok.

amarax:
@drewjhart we can use the circle pagination indicators for now, but they're not ideal because the hit area would be too small for mobile usage. 

In theory, the following CSS should do the trick:
```css
.swiper-pagination-bullet {
  width: 20px;
  height: 6px;
  background-color: #999;
  border-radius: 0;
}
```

You can consider applying this to a Carousel wrapper class around the SwiperJS component that we can use throughout RightOn.

I understand if the proposed fix above doesn't fit with the rest of how the components are built, and the actual solution may be more complex.

This is a low priority UX issue, and I think the functionality we have for now is sufficient!

drewjhart:
> @drewjhart we can use the circle pagination indicators for now, but they're not ideal because the hit area would be too small for mobile usage.
> 
> In theory, the following CSS should do the trick:
> 
> ```css
> .swiper-pagination-bullet {
>   width: 20px;
>   height: 6px;
>   background-color: #999;
>   border-radius: 0;
> }
> ```
> 
> You can consider applying this to a Carousel wrapper class around the SwiperJS component that we can use throughout RightOn.
> 
> I understand if the proposed fix above doesn't fit with the rest of how the components are built, and the actual solution may be more complex.
> 
> This is a low priority UX issue, and I think the functionality we have for now is sufficient!

@amarax yeah, that css seems solid but the complexity comes from accessing the pagination bullets in the broader context of the page. I get your point about needing enough surface area for an interaction on mobile though - seems like we're going to need to get these to match the Figma. I'll take a closer look at this tomorrow morning and see what I can do.

drewjhart:
> > @drewjhart we can use the circle pagination indicators for now, but they're not ideal because the hit area would be too small for mobile usage.
> > In theory, the following CSS should do the trick:
> > ```css
> > .swiper-pagination-bullet {
> >   width: 20px;
> >   height: 6px;
> >   background-color: #999;
> >   border-radius: 0;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can consider applying this to a Carousel wrapper class around the SwiperJS component that we can use throughout RightOn.
> > I understand if the proposed fix above doesn't fit with the rest of how the components are built, and the actual solution may be more complex.
> > This is a low priority UX issue, and I think the functionality we have for now is sufficient!
> 
> @amarax yeah, that css seems solid but the complexity comes from accessing the pagination bullets in the broader context of the page/our styling system. I get your point about needing enough surface area for an interaction on mobile though - seems like we're going to need to get these to match the Figma. I'll take a closer look at this tomorrow morning and see what I can do.

hey @amarax, just to follow up I found a way to get those pagination bullets shaped right. check [play.rightoneducation.com](play.rightoneducation.com) for the update!

play/package.json
```diff
@@ -98,5 +101,7 @@
     "@storybook/react-docgen-typescript-plugin": "1.0.6--canary.9.cd77847.0",
     "styled-components": "^5",
     "@mui/styled-engine": "npm:@mui/styled-engine-sc@latest"
-  }
+  },
+  "readme": "ERROR: No README data found!",
```
ryanabooth, line 105: 😕

play/src/components/AnswerSelector.tsx
```diff
@@ -84,19 +85,23 @@ export default function AnswerSelector({
...
-          right: isSubmitted ? '17px' : '16px',
-          width: '16px',
-          height: '16px',
-          paddingTop: '3px', // 16px + 3 px + 3px = 22px -> same line height as text, all per Figma
-          paddingBottom: '3px',
+          right: isSubmitted
+            ? `${theme.sizing.smallPadding + 1}px`
+            : `${theme.sizing.smallPadding}px`,
+          width: `${theme.sizing.smallPadding}px`,
+          height: `${theme.sizing.smallPadding}px`,
```
ryanabooth, line 103: You will probably want to avoid using sizing variables intended for centralized padding for the size of things like images. You can create other sizing variables, or consider whether this is a standardized size at all.

play/src/components/ButtonSubmitAnswer.tsx
```diff
@@ -2,25 +2,25 @@ import React from 'react';
...
-  background: '#909090',
+const AnswerButtonDisabled = styled(AnswerButton)(({ theme }) => ({
+  background: `${theme.palette.primary.extraDarkGrey}`,
   boxShadow: 'none',
   '&:hover': {
-    background: '#909090',
+    background: `${theme.palette.primary.extraDarkGrey}`,
   },
-});
+}));
```
ryanabooth, line 23: great use of theme variables

play/src/components/CardAnswer.tsx
```diff
@@ -72,32 +53,33 @@ export default function CardAnswer({
...
+            />
+          ))}
+        </Stack>
+        <ButtonSubmitAnswer
+          isSubmitted={isSubmitted}
+          handleSubmitAnswer={handleSubmitAnswer}
+          isSelected={!isNullOrUndefined(selectedAnswer)}
+        />
+      </BodyCardContainer>
     </BodyCard>
```
ryanabooth, line 83: Very readable code.

play/src/components/CardQuestion.tsx
```diff
@@ -0,0 +1,35 @@
...
+  BodyCard,
+  BodyCardContainer,
+} from '../lib/styledcomponents/StyledComponents';
+
+interface CardQuestionProps {
+  questionText: string[];
+  imageUrl: string;
+}
+
+export default function CardQuestion({
```
ryanabooth, line 13: QuestionCard?

play/src/components/ScoreIndicator.stories.tsx
```diff
@@ -18,32 +18,8 @@ const Template: ComponentStory<typeof ScoreIndicator> =
...
-LowScore.args = {
-  score: 1,
-  newPoints: 0,
-};
-
-export const NoScore = Template.bind({});
-NoScore.args = {
-  score: null,
-  newPoints: 0,
-};
```
ryanabooth, line 49: If these cases are useful to view in the storybook, why not keep them?

play/src/lib/Theme.tsx
```diff
@@ -1,15 +1,23 @@
 import { createTheme } from '@mui/material/styles';
```
ryanabooth, line 1: 👍

ryanabooth (approved):

play/package.json
drewjhart: resolved

play/src/components/AnswerSelector.tsx
drewjhart: makes a lot of sense, resolved to original values for now

play/src/components/ButtonSubmitAnswer.stories.tsx
```diff
@@ -18,20 +18,8 @@ const Template: ComponentStory<typeof ButtonSubmitAnswer> =
...
-AnswerSelected.args = {
-  isSelected: true,
-  isSubmitted: false,
-};
-
-export const AnswerSubmitted = Template.bind({});
-AnswerSubmitted.args = {
-  isSelected: true,
-  isSubmitted: true,
-};
```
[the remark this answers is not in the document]
drewjhart: This is the result of some back and forth with U/X. Yong Lin has requested that for components that are molecules or smaller that we remove the stories and allow him to just manipulate the props with the controls below the UI that storybook offers. He thinks that will be simpler than having individual stories for each interaction at this scale.

play/src/components/CardQuestion.tsx
drewjhart: much better, changed to questioncard and cardanswer has been changed to answercard.

play/src/components/ScoreIndicator.stories.tsx
drewjhart: Also a result of the U/X coordination with Yong Lin. It's not a huge lift to include these cases but I'm ok simplifying the base level components in this way if it's helpful to his workflow (I can still QA the interactions I need to through the controls storybook provides).

play/src/components/Timer.tsx
```diff
@@ -3,22 +3,22 @@ import { styled } from '@mui/material/styles';
...
   alignItems: 'center',
-  marginLeft: '24px',
-  marginRight: '24px',
-  marginTop: '8px',
-  marginBottom: '8px',
-});
+  marginLeft: `${theme.sizing.mediumPadding}px`,
+  marginRight: `${theme.sizing.mediumPadding}px`,
+  marginTop: `${theme.sizing.extraSmallPadding}px`,
+  marginBottom: `${theme.sizing.extraSmallPadding}px`,
```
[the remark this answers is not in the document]
drewjhart: Really good idea. I've just been parsing the Figma design to extract these values so I'll be going through it with U/X on Monday. I'll bring up the gutters at that point to see about standardising this further.

play/src/components/Timer.tsx
```diff
@@ -80,6 +84,11 @@ export default function Timer({
     return () => cancelAnimationFrame(animationRef.current ?? 0);
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
+  // Update the isPausedRef when the isPaused prop changes
+  useEffect(() => {
+    isPausedRef.current = isPaused;
+  }, [isPaused]); // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
drewjhart: Yeah, this is rough tbh. The more I try to make `requestAnimationFrame` work, the more complicated it gets. The issue this is trying to resolve is that as soon as we call the recursive function, it holds a stale value for `isPaused` and then can't break out of the recursion if the prop changes. The `useEffect` updates a ref that we can then reference in the function to access the most up-to-date value. 

It's clunky and I've got two `useEffects` and bunch of `refs` for something that should be simple. I'll book some time with you next week and, in the meantime, will probably circle right back around to using `setinterval` to do this in a way that doesn't hurt to write hahah. Would be great to touch base on it next week though

PR #588, opened 2023-04-19
[Play] - Enter Game Code Screen

drewjhart:
This PR adds the `EnterGameCode` screen to the flow below:

Splash Screen -> Enter Game Code -> Enter Player Name -> Select Avatar -> How to Play Carousel

The Figma reference is here:
image
https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?node-id=13554-169657&t=nuq6BdMbpGPDYcaa-4

PR Structure:

[[Play] - Intro Splash Screen and Join Game Flow Base](https://github.com/rightoneducation/righton-app/pull/586)
-> [[Play] - Enter Game Code Screen #588](https://github.com/rightoneducation/righton-app/pull/588)
->->  [[Play] - Enter Player Name Screen #589](https://github.com/rightoneducation/righton-app/pull/589)
->->->  [[Play] - Select Avatar Page #590](https://github.com/rightoneducation/righton-app/pull/590)
->->->-> [[Play] - How to Play Screen #591](https://github.com/rightoneducation/righton-app/pull/591)

PR #586, opened 2023-04-15
[Play] - Intro Splash Screen

drewjhart:
This PR acts as the base for the `JoinGame` flow that is as follows:

Splash Screen -> Enter Game Code -> Enter Player Name -> Select Avatar -> How to Play Carousel

Additionally, it adds the first screen, `SplashScreen.tsx` in that flow. The Figma reference is here:
image
https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?node-id=13583-154467&t=nuq6BdMbpGPDYcaa-4

Code Organization: 
The main idea here is to create a `JoinGame` page that has a switch to flip between the different Join Game states. I'd rather do this than hoist the whole thing to `GameSessionContainer` because the states for Join Game are distinct from those in `GameSessionState.current.` `GameSessionState.current` just has `TEAMS.JOINING` and each of the Join Game states are a subset of that one state. I'd like to reflect this in our code for clarity.

PR Structure:
To be honest, I am reverse engineering this PR. It used to have every screen integrated and I've broken it down on a screen-by-screen basis for ease of review. The structure is now:

[[Play] - Intro Splash Screen and Join Game Flow Base](https://github.com/rightoneducation/righton-app/pull/586)
-> [[Play] - Enter Game Code Screen #588](https://github.com/rightoneducation/righton-app/pull/588)
->->  [[Play] - Enter Player Name Screen #589](https://github.com/rightoneducation/righton-app/pull/589)
->->->  [[Play] - Select Avatar Page #590](https://github.com/rightoneducation/righton-app/pull/590)
->->->-> [[Play] - How to Play Screen #591](https://github.com/rightoneducation/righton-app/pull/591)

_Note: this PR contains image files that aren't directly associated with the Splash Screen (it contains monster icon images for the Select Avatar page). This is because I've developed an object to handle these icons that's also shared by the footer container. I didn't want to undo and then reapply these changes so hopefully this isn't confusing. I've tagged where this occurs in each of the files with comments._

PR #589, opened 2023-04-19
[Play] - Enter Player Name Screen

drewjhart:
This PR adds the `EnterPlayerName` screen to the flow below:

Splash Screen -> Enter Game Code -> Enter Player Name -> Select Avatar -> How to Play Carousel

The Figma reference is here:
image
https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?node-id=13554-169658&t=nuq6BdMbpGPDYcaa-4

PR Structure:

[[Play] - Intro Splash Screen and Join Game Flow Base](https://github.com/rightoneducation/righton-app/pull/586)
-> [[Play] - Enter Game Code Screen #588](https://github.com/rightoneducation/righton-app/pull/588)
->->  [[Play] - Enter Player Name Screen #589](https://github.com/rightoneducation/righton-app/pull/589)
->->->  [[Play] - Select Avatar Page #590](https://github.com/rightoneducation/righton-app/pull/590)
->->->-> [[Play] - How to Play Screen #591](https://github.com/rightoneducation/righton-app/pull/591)


play/src/components/joingame/EnterPlayerName.tsx
```diff
@@ -0,0 +1,146 @@
...
+            <Grid item xs={6}>
+              <IntroTextField
+                fullWidth
+                variant="filled"
+                autoComplete="off"
+                onChange={(newValue) => {
+                  handlePlayerFirstNameChange(newValue.target.value);
+                }}
+                onFocus={(newValue) => {
+                  if (newValue.target.value === 'First Name') {
```
maniramezan, line 65: Same comment similar to the other PR, probably wanna extract `'First Name'` into a const and then use that instead of hardcoding string across the file

maniramezan: Also, same comment regarding `placeholder` usage, looks like we use `onFocus` to remove the default value in the field while if we use `placeholder` we shouldn't need this.

And if we do this, then you can use something like `input::placeholder` or there should be a similar approach to modify the text color for placeholder vs. normal message

drewjhart: placeholder values have been moved to an enum in `/lib/PlayModels` and placeholder values are fixed per changes in HowToPlay.

PR #590, opened 2023-04-19
[Play] - Select Avatar Page

drewjhart:


play/src/components/joingame/SelectAvatar.tsx
```diff
@@ -0,0 +1,185 @@
...
+                  alt="avatar"
+                  sx={{ borderWidth: index === selectedAvatar ? '6px' : '0px' }}
+                />
+              </IconContainer>
+            ))}
+          </GridContainer>
+        </Stack>
+        <MonsterContainer sx={{ height: isMobileDevice ? '250px' : '300px' }}>
+          <Monster
+            src={monsterMap[selectedAvatar || 0].monster}
```
maniramezan, line 171: Since we're defaulting it to `0` here, should this property just not be nullable and has a default value of `0`?

drewjhart: I'm trying to double up the use of selectedAvatar here (for both the monster image and the monster icon) but there's an edge case this is catching. When a player first navigates to the screen, selectedAvatar is null because I don't think we want any selection on the avatar image. However, the main monster image should still be at `0` so that a default is selected (we ran into crashes with the `mobile` app when we didn't do this). The condition is below:

image

This is in contrast to when a player selects the first monster. In that case selectedAvatar goes to 0, and the white border does appear:

image

So I'd like to maintain a default of null (to prevent the white border from appearing) while also providing a default value here so a monster always shows. I think this is working intended but it's definitely not super clear. I've added a comment that explains this line in the code and if you have another way of improving the readability, I'm all ears.

maniramezan: I think there should be a correlation between monster photos and the main one shown below. If we wanna go with `null`, then probably we shouldn't also show the large monster below. It's a bit confusing if initial load has no monsters selected, but shows the first monster image. Especially, if we use `null`, we should disable `choose` button so user cannot move on to the next screen without choosing an avatar.

drewjhart: Yeah, touche. I'm off tomorrow and Friday but let me talk to Yong Lin about how the flow should work so we can produce a better correlation. I'll follow up with a PR next week.

PR #591, opened 2023-04-19
[Play] - How to Play Screen

drewjhart:
This PR adds the `HowToPlay` screen to the flow below:

Splash Screen -> Enter Game Code -> Enter Player Name -> Select Avatar -> How to Play Carousel

The Figma reference is here:
image
https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?node-id=13554-169660&t=nuq6BdMbpGPDYcaa-4

PR Structure:

[[Play] - Intro Splash Screen and Join Game Flow Base](https://github.com/rightoneducation/righton-app/pull/586)
-> [[Play] - Enter Game Code Screen #588](https://github.com/rightoneducation/righton-app/pull/588)
->->  [[Play] - Enter Player Name Screen #589](https://github.com/rightoneducation/righton-app/pull/589)
->->->  [[Play] - Select Avatar Page #590](https://github.com/rightoneducation/righton-app/pull/590)
->->->-> [[Play] - How to Play Screen #591](https://github.com/rightoneducation/righton-app/pull/591)


play/src/components/joingame/SelectAvatar.tsx
```diff
@@ -113,34 +77,36 @@ const Monster = styled('img')({
...
 }: SelectAvatarProps) {
   const theme = useTheme();
 
   return (
-    <JoinGameBackgroundContainer>
-      <StackContainer>
-        <Stack spacing={2}>
+    <JoinGameBackgroundContainer >
+      <StackContainer isMobileDevice={isMobileDevice} >
+        <Stack spacing={2} >
```
maniramezan, line 109: nit: extra space before `>`

play/src/components/joingame/SplashScreen.tsx
```diff
@@ -1,24 +1,46 @@
...
+})
+);
+
 export default function SplashScreen() {
   const theme = useTheme();
   return (
+
     <JoinGameBackgroundContainer>
+        <HeroContainer>
       <StackContainer spacing={5}>
```
maniramezan, line 42: nit: Indent for this component and subcomponents

play/src/components/joingame/SplashScreen.tsx
```diff
@@ -28,15 +50,23 @@ export default function SplashScreen() {
           src={Logo}
           alt="Question"
         />
-        <Typography variant="h2" sx={{ weight: 700, textAlign: 'center' }}>
+        <Typography variant="h2" sx={{ weight: 700, textAlign: 'center', paddingLeft: `${theme.sizing.mediumPadding}px`, paddingRight: `${theme.sizing.mediumPadding}px`  }}>
           Unlocking every student`s potential in math!
```
maniramezan, line 54: nit: Should this be `'` or `` ` ``

play/src/containers/GameSessionContainer.tsx
```diff
@@ -16,15 +16,81 @@ export default function GameSessionContainer() {
...
+      case 3:
+        setjoinGameState(JoinGameState.SELECTAVATAR);
+        break;
+      case 4:
+        setjoinGameState(JoinGameState.HOWTOPLAY);
+        break;
+      default:
+        // Handle invalid value here
+        break;
+    }
```
maniramezan, line 43: In Typescript, you can assign values to enums, any reason we don't use that approach instead of having enum and then a separate value set for each?

play/src/containers/GameSessionContainer.tsx
```diff
@@ -16,15 +16,81 @@ export default function GameSessionContainer() {
...
+            checked={joinGameState === JoinGameState.SELECTAVATAR}
+            onChange={handleOptionChange}
+          />
+          <input
+            type="radio"
+            name="option"
+            value="4"
+            checked={joinGameState === JoinGameState.HOWTOPLAY}
+            onChange={handleOptionChange}
+          />
```
maniramezan, line 89: You might be able to reduce this boilerplate code by iterating over `JoinGameState` enum cases and create the `input` for each

play/src/lib/styledcomponents/StyledComponents.tsx
```diff
@@ -97,12 +97,46 @@ export const IntroTextField = styled(TextField)(({ theme }) => ({
...
   display: 'flex',
   alignItems: 'center',
 }));
+
+// icons for the avatar select screen
+interface AvatarIconProps {
+  isSelected: boolean;
+}
+
+export const AvatarIcon = styled('img', {
```
maniramezan, line 116: General question on this file, any reason we're adding all in one file instead of breaking it into separate files?

play/src/lib/styledcomponents/SC_IntroButton.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Template: ComponentStory<typeof IntroButton> =
+  function ButtonSubmitAnswerTemplate(args) {
+    return (
+      <ThemeProvider theme={Theme}>
+        <IntroButton {...args}>
+        <Typography variant="h2" sx={{ textAlign: 'center' }}>
+            Join Game
+          </Typography>
+          </IntroButton>
```
maniramezan, line 21: nit: indent

play/src/components/joingame/EnterGameCode.tsx
```diff
@@ -61,6 +61,11 @@ export default function EnterGameCode({
                 handleGameCodeChange('');
               }
             }}
+            onBlur={(newValue) => {
+              if (newValue.target.value === '') {
+                handleGameCodeChange('####');
+              }
```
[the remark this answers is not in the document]
drewjhart: Thanks for the review on this, Mani. 

Good call. The PR's have gotten messy here but this is fixed in [PR588 - Enter Game Code](https://github.com/rightoneducation/righton-app/pull/588/files#diff-d3f9f77923ecf10b846ae11b28b0571f17fb37366a85b569730601ade2dd5c5a) based on Ryan's comment. When I merge the nested PR's in, that correct version will take precedent.

play/src/components/joingame/EnterPlayerName.tsx
```diff
@@ -21,24 +24,28 @@ const PaddedContainer = styled(Box)(({ theme }) => ({
...
+  isMobileDevice,
   inputError,
 }: EnterPlayerNameProps) {
   const theme = useTheme();
+
   return (
     <JoinGameBackgroundContainer>
-      <StackContainer spacing={5}>
+      
+      <StackContainer spacing={isMobileDevice ? 2 : 5}>
```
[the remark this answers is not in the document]
drewjhart: A 'mobile device' here is anything under 700px. Our Figma design includes desktop and 'mobile' designs, and then the responsive design handles tablet conditions:

image

This is differentiated is based on our breakpoint settings (sm=700px). 
Currently, the media query handles whether or not the answers are displayed in two columns or one and if a avatar on the avatar select screen is scaled down a bit. 

Overall though, I think `isMobileDevice` is a pretty ambiguous label. I've renamed it to `isSmallDevice` because that's as explicitly accurate as possible (it is small with respect to our breakpoints). If you've got a better suggestion though, would definitely appreciate it!

play/src/lib/styledcomponents/StyledComponents.tsx
drewjhart: The idea on this file is that it acts as a library for the different styles. So I am thinking about it more as though it's like a css page instead of a series of components. That being said, as I write more pages it's bloating a lot and definitely needs to be broken down. 

Do you think it would make more sense to extract each component here (similar to a standard React structure) or should they be grouped by use cases (inputs, layout items, etc)? I think it might make sense to just extract each component (they're called styled components after all) but it will grow the number of files dramatically.

maniramezan: Hmm, I think it depends on how large this file becomes and how many more styles we're going to add. Having too many files wouldn't be that bad as we can always use VSCode to just jump to a file and if we follow the pattern similar to ReactJS, that means by default, you know that `BodyCard` style is in `BodyCard` under `StyledComponents`. If we go this path, you probably want to start adding `Style` suffix to all of these. This way, it's obvious it's just CSS styling and not a component. This way, you can have `BodyCard` reactjs component and `BodyCardStyle`  for CSS part

play/src/components/joingame/EnterPlayerName.tsx
```diff
@@ -54,64 +61,46 @@ export default function EnterPlayerName({
...
+              inputProps: {
+                style: {
+                  color: theme.palette.primary.darkBlue,
+                  paddingTop: '9px',
+                  textAlign: 'center',
+                  fontSize: `${theme.typography.h2.fontSize}px`,
+                },
+              },
+            }}
+          />
```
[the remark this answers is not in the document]
drewjhart: fixed via prettier

play/src/components/joingame/EnterPlayerName.tsx
```diff
@@ -54,64 +61,46 @@ export default function EnterPlayerName({
...
```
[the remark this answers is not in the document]
drewjhart: fixed via prettier

play/src/components/joingame/SplashScreen.tsx
drewjhart: fixed  ('center' has gone to `center`)

play/src/containers/GameSessionContainer.tsx
drewjhart: Fair point. this was just the debug radio controls for review though so I've removed all the code from the PR.

play/src/containers/GameSessionContainer.tsx
drewjhart: Same as above, lazy debug controls pulled out of PR

play/src/lib/styledcomponents/StyledComponents.tsx
drewjhart: Yeah, agreed. I switched over to this now and it seems a lot cleaner. One note, instead of `Style` I've used `Styled` as we end up using the styled component like `<StyledComponent />`. If we went with `Style` we would have `<ComponentStyle />` which felt kind of odd to me, so I'm doing `<ComponentStyled />` instead. Still feels clunky but communicates that it's a styled component at least. Let me know if you have a better way of doing this though

play/src/lib/styledcomponents/SC_IntroButton.stories.tsx
drewjhart: fixed via prettier

play/src/lib/styledcomponents/StyledComponents.tsx
maniramezan: Sounds good thanks! I think as you're working mainly on this, see how it works out and if still didn't feel good or too much work to do it this way, we can revisit and find alternatives

PR #597, opened 2023-05-04
[Networking] [DebugPanel] Update to Yarn and update all packages

maniramezan:
- Update to yarn instead of npm for networking and debug panel
- Update the dependencies to the latest version


networking/.gitignore
drewjhart, line 1: this is just for my personal curiosity, but is this just a safeguard in case people have old copies of package-lock.json in their environments? moving forward with your changes, we'd expect them not to have a package-lock.json and instead a yarn.lock right?

debug-panel/package.json
drewjhart, line 1: I'd like to follow up on the procedure for this with our other apps, as I think we have a lot of older packages (the website for instance has a bunch of deprecated packages which could cause issues).

maniramezan: Yeah definitely, there's a package called [npm-check-updates](https://www.npmjs.com/package/npm-check-updates) that I've installed locally. It helps to update all the packages to the latest easily

networking/.gitignore
maniramezan: Yeah, so when you want to add an existing file to `.gitignore` you need to delete it and then add the file to the ignore list. This way, others with local `package-lock.json`, it never shows up in the git changes. And even is they use `npm install` instead of `yarn install`, it still doesn't matter as the lock file would never be checked it.

PR #599, opened 2023-05-05
[Play] - Pregame Countdown

drewjhart:
This PR adds the pregame countdown page that occurs when the teacher moves from `GameSessionState.TEAMS_JOINING` to `GameSessionState.CHOOSE_CORRECT_ANSWER`

**Screenshot:**
Screenshot 2023-05-05 at 4 09 50 PM

**Figma Reference:**
image
[https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?node-id=13554-169661&t=PGzsP6wda24aGPzP-4](**url**)


play/src/components/RadialTimer.tsx
maniramezan, line 1: Might worth to have some comments or a link if this is inspired by any posts online

play/src/containers/GameSessionContainer.tsx
```diff
@@ -14,7 +15,8 @@ import StartPhase2 from '../pages/StartPhase2';
 import { JoinGameState, FinalResultsState } from '../lib/PlayModels';
 
 export default function GameSessionContainer() {
-  const [gameSession, setGameSession] = useState(   // eslint-disable-line @typescript-eslint/no-unused-vars
+  const [gameSession, setGameSession] = useState(
+    // eslint-disable-line @typescript-eslint/no-unused-vars
```
maniramezan, line 19: If this is temporarily and needs to be clean up, one common practice is adding `// TODO: Remove this` or something with more info so you can easily find them and make sure these are resolved before final release

play/src/lib/Theme.tsx
```diff
@@ -100,6 +102,7 @@ declare module '@mui/material/styles' {
     darkGrey?: string;
     lightGrey?: string;
     correctColor?: string;
+    countdownColor?: string;
```
maniramezan, line 105: Any reason for this be nullable?

maniramezan (approved):

play/src/components/RadialTimer.tsx
drewjhart: Thanks for the reviews on this PR and the others. Appreciate it as always.

And added reference comment back in (this is reused code from host/mobile)

play/src/containers/GameSessionContainer.tsx
drewjhart: Added ``// TODO`` statements to explain this line and the lines below it

play/src/lib/Theme.tsx
drewjhart: Good point. Removed `?`

PR #617, opened 2023-05-19
[Play] - Container Refactors

drewjhart:
**Issue:**
The current structuring of our code is overly complicated, with three containers and a condition in which `gameSession` is initialised with `null`, forcing us to do a bunch of null checks all the way down our code. 

**Resolution:**
After speaking with Mani and Ryan, we came up with some requirements for the new structure:

Mani's recommendations:
1. Use only two containers
2. Eliminate the need for `null` initialisation of `gameSession` object

Ryan's recommendations:
1. use routing to split the app into something like `/` for `preGame` and `/game` for `gameInProgress`
2. use either route param or localStorage to transition between the two
3. use a secondary container to handle the api calls
4. perform null logic (using loading/error/data states) in that secondary container to mitigate the need for `null` checks in child containers

The included code implements these recommendations. Here are the important components:
1. `PregameContainer.tsx` - handles all the joining game functionality (enter gamecode etc). When user selects avatar, `team` is added to `gameSession` object on the backend and data is stored in `localStorage`.
2.  `useFetchLocalData.tsx` - custom hook for fetching local game session data. Exported so we can use a `navigate('/')` without compromising rules of hooks.
3. `useFetchandSubscribeGameSession.tsx` - custom hook for fetching the current game session and subscribing to it. Returns an object formatted as `{ isLoading, error, gameSession}` to track game loading, error state and data populating.
4. `GameSessionContainer.tsx` - retrieves local data and subscribes to game via custom hooks. Renders appropriate components based on response from subscription.
5. `GameSessionSwitch.tsx` - holds all game pages and switches them based on `gameSessionState`. This is the normal Game In Progress flow. 

**Preview:**
No errors, player joins game, game loads, player waits in lobby until teacher starts game:

https://github.com/rightoneducation/righton-app/assets/79417944/626c5e2c-b961-4854-80fe-41ba667b4c18

With error, modal pops up, player attempts to rejoin, player quits out to main menu:
New video without `useEffect` (june 1, 2023):

https://github.com/rightoneducation/righton-app/assets/79417944/d56a53cd-6990-4feb-b5ed-1e2bb798c34d

Old video with janky `useEffect` (may 31 2023):

https://github.com/rightoneducation/righton-app/assets/79417944/6af25eb2-b99c-4e30-aa86-1090aac40f5b


play/src/containers/GameInProgressContainer.tsx
```diff
@@ -1,118 +1,60 @@
...
-          teamId={teamId}
-          score={score}
-          addTeamAnswerToTeamMember={addTeamAnswerToTeamMember}
-        />
-      );
-    case GameSessionState.CHOOSE_TRICKIEST_ANSWER:
-    case GameSessionState.PHASE_1_DISCUSS:
-    case GameSessionState.PHASE_2_DISCUSS:
+  // this listens for changes to the retry state and updates the error text accordingly
+  React.useEffect(() => {
```
drewjhart, line 37: this feels janky to me. I'm using it to listen for updates on the retry/subscription retry so that when the user retries, I can update the modal text from 'An error has occurred' to 'Retrying' and then back to 'An error has occurred' if the custom hook fails again. I feel like I'm putting state logic where it shouldn't go but I can't find another place to put it (originally it was down in the modal but I was getting a bunch of stale data).

play/src/containers/GameInProgressContainer.tsx
```diff
@@ -1,118 +1,60 @@
...
-          {...gameSession}
-          currentState={currentState}
-          score={score}
-          selectedAvatar={teamAvatar}
-          teamId={teamId}
-          leader={leader}
-        />
+        <>
+          <HowToPlay isLoading={false} isError />
+          <AlertModal errorText={errorText} retry={retry} handleRetry={handleRetry}/>
```
drewjhart, line 48: I think you mentioned there might be an alternate way to do this (without passing down the booleans as props into the components themselves). Is there a cleaner approach than this?

play/src/containers/GameInProgressContainer.tsx
drewjhart: I've updated the code to remove this `useEffect` through the moving of all the text handling to the `useFetchAndSubscribeGameSession` hook. I think this is a lot cleaner than having to listen for the changes again here and the behaviour of the modal is nicer (it doesn't render on the retry click and then appears again when the error is thrown). I've updated the video so you can see it in action there.

play/src/containers/GameInProgressContainer.tsx
drewjhart: cleaned up the props here so it's just a single `mode` enum that changes based on the received status from the custom hook

play/src/pages/pregame/SelectAvatar.tsx
```diff
@@ -136,7 +137,13 @@ export default function SelectAvatar({
           <Typography variant="h2" sx={{ textAlign: 'center' }}>
             {`${firstName} ${lastName}`}
           </Typography>
-          <GamePlayButtonStyled onClick={handleAvatarSelectClick}>
+          <GamePlayButtonStyled
+            onClick={() => {
+              handleAvatarSelectClick();
+              setIsButtonPressed(true);
```
maniramezan, line 143: Is `handleAvatarSelectClick` async? Wondering if `setIsButtonPressed` should be called before this or not

play/src/hooks/useFetchAndSubscribeGameSession.tsx
```diff
@@ -0,0 +1,79 @@
...
+  useEffect(() => {
+    // prevents runaway condition by ignoring updates to state after component unmounts
+    let ignore = false;
+    // if player is retrying after an error, reset state values to 'isLoading' conditions unless the api call fails again
+    if (retry > 0) {
+      setIsLoading(true);
+      setError('');
+    }
+    // if gamesessionId is null, set error and prevent api calls
+    if (isNullOrUndefined(gameSessionId)){
```
maniramezan, line 31: nit: space before `{`

play/src/hooks/useFetchAndSubscribeGameSession.tsx
```diff
@@ -0,0 +1,79 @@
...
+      setError(`${t('error.connecting.gamesessionerror')}`);
+      setIsLoading(false);
+    }
+    else {
+      apiClient
+        .getGameSession(gameSessionId)
+        .then((fetchedGame) => {
+          if (!fetchedGame)
+            setError(`${t('error.connecting.gamesessionerror')}`);
```
maniramezan, line 40: Should also have `return`?

play/src/hooks/useFetchAndSubscribeGameSession.tsx
```diff
@@ -0,0 +1,79 @@
...
+              fetchedGame.id,
+              (response) => {
+                if (!response)
+                  setError(`${t('error.connecting.subscriptionerror')}`);
+                // Update the gameSession object and trigger the callback
+                if (!ignore)
+                  setGameSession((prevGame) => ({ ...prevGame, ...response }));
+              }
+            );
+            setIsLoading(false);
```
maniramezan, line 52: This should probably be the first statement here

play/src/hooks/useFetchAndSubscribeGameSession.tsx
```diff
@@ -0,0 +1,79 @@
...
+            setError(`${t('error.connecting.gamesessionerror')}`);
+          try {
+            const gameSessionSubscription = apiClient.subscribeUpdateGameSession(
+              fetchedGame.id,
+              (response) => {
+                if (!response)
+                  setError(`${t('error.connecting.subscriptionerror')}`);
+                // Update the gameSession object and trigger the callback
+                if (!ignore)
+                  setGameSession((prevGame) => ({ ...prevGame, ...response }));
```
maniramezan, line 49: same as above, does it need `return`? otherwise it continues the code after the first condition

play/src/hooks/useFetchAndSubscribeGameSession.tsx
```diff
@@ -0,0 +1,79 @@
...
+              ignore = true;
+              gameSessionSubscription.unsubscribe();
+            };
+          } catch (e) {
+            setIsLoading(false);
+            if (e instanceof Error)
+              setError(e.message);
+            else
+              setError(`${t('error.connecting.subscriptionerror')}`);
+          }
```
maniramezan, line 63: Not an expert in promises on js, but wondering if the higher level `catch`  also captures inner error handling, might worth to just throw the error here and let the outer catch unifies the error handling

play/src/lib/PlayModels.tsx
```diff
@@ -35,25 +35,36 @@ export enum AnswerState {
...
- * JoinGameState handles the substate changes for GameSessionState.JOIN_GAME
- * @enum JoingGameState
+ * OverallGameState handles the state changes between the Pregame or GameInProgress phases
+ * Pregame: user-controlled, joining the game
+ * GameInProgress: subscription-controlled, playing the game
+ * @enum OverallGameState
+ */
+export enum OverallGameState {
+  PREGAME = 'PREGAME',
+  GAME_IN_PROGRESS = 'GAME_IN_PROGRESS',
```
maniramezan, line 45: You don't need to define string value always if you're not using it. Same for `HowToPlayMode`. If you don't need an absolute value for each case, you can omit defining it

play/src/pages/pregame/HowToPlay.tsx
```diff
@@ -35,69 +36,82 @@ const HowToPlaySwiper = styled(Swiper)({
...
+          </SwiperSlide>
+          <SwiperSlide>
+            <HowToPlaySlide2Content />
+          </SwiperSlide>
+          <SwiperSlide>
+            <HowToPlaySlide3Content />
+          </SwiperSlide>
+          <SwiperSlide>
+            <HowToPlaySlide4Content />
+          </SwiperSlide>
```
maniramezan, line 95: hmm, maybe have an array that includes all the `HowToPlaySlideXContent` components and then use `map` to create this. Not important, but can be a good practice to reduce the code while making it easier to add any new content as instead of copy pasting code here, you just need to add the new content in the array

maniramezan (approved):
Nice job on the changes, it looks cleaner even in the PR, hopefully it has reduced some complexity in development as well

play/src/components/GameSessionSwitch.tsx
```diff
@@ -0,0 +1,139 @@
...
+  gameSession,
+  pregameModel,
+}: GameInProgressContainerProps) {
+  const [isPregameCountdown, setIsPregameCountdown] = useState<boolean>(true);
+  const { currentState } = gameSession;
+  const currentQuestion =
+    gameSession.questions[gameSession.currentQuestionIndex ?? 0];
+  const currentTeam = gameSession.teams?.find(
+    (team) => team.id === pregameModel.teamId
+  );
```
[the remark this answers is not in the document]
drewjhart: Totally agree with this. How do you mean force unwrap these? I'm doing these kind of messy checks because the IGameSession interface has an optional flag on currentQuestionIndex and teams. From a `play` perspective, this isn't relevant (as they are both always going to be present) but from a `host` and `central` perspective they can be `null`. 

Is there a better way of handling this condition?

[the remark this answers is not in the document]
maniramezan: Sorry I'm using Swift terminologies here, not sure what it's called in TS, using `!` instead of `?` in here. So generally, we want to have only one source to check for nullability or enforce it here. I'm also thinking if it makes sense to have a different models or annotations here and handle these in networking, like if it's host app, it returns nullable and if it's play, it returns non-nullable.

We might be able to add helper methods to `IGameSession` as well something like:

```typescript
get nonNullTeams(): Array<ITeam> {
    if (isNullOrUndefined(teams) {
        console.error("Teams can't be null")
        return []
    }
    return teams // Typescript should be smart enough to know this is neither `null` nor `undefined`
}
```

play/src/components/GameSessionSwitch.tsx
```diff
@@ -0,0 +1,139 @@
...
+  // locally held score value for duration of gameSession, updates backend during each PHASE_X_RESULTS
+  const [score, setScore] = useState(currentTeam?.score ?? 0);
+  const leader = true;
+  const answerChoices =
+    currentQuestion?.choices!.map((choice: IChoice) => ({  // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
drewjhart: Good call, resolved.

play/src/components/GameSessionSwitch.tsx
```diff
@@ -0,0 +1,139 @@
...
+    try {
+      await apiClient.addTeamAnswer(
+        pregameModel.teamMemberId,
+        question.id,
+        answerText,
+        gameSessionState === GameSessionState.CHOOSE_CORRECT_ANSWER,
+        gameSessionState !== GameSessionState.CHOOSE_CORRECT_ANSWER
+      );
+    } catch (error) {
+      console.error(error);
```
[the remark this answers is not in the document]
drewjhart: I really appreciate this comment. I'm going to meet with U/X this afternoon to talk about how we want to handle errors and then I'm going to do that error handling in a separate PR (I think the scoring and the answering error handling here will need some thought in terms of how any 'retry' mechanics are handled). I tried to write a generic 'retry' function at this level in the component tree and it got pretty messy as I was storing the initial team answer and score data and then trying to reference it. 

This is all to say that I'll coordinate with U/X and then create a new PR for better error handling here.

play/src/components/GameSessionSwitch.tsx
```diff
@@ -0,0 +1,139 @@
...
+          addTeamAnswerToTeamMember={addTeamAnswerToTeamMember}
+        />
+      );
+    case GameSessionState.PHASE_1_RESULTS:
+    case GameSessionState.PHASE_2_RESULTS:
+      return (
+        <PhaseResults
+          {...gameSession}
+          gameSession={gameSession}
+          currentQuestionIndex={gameSession!.currentQuestionIndex ?? 0} // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
drewjhart: Corrected.

play/src/components/GameSessionSwitch.tsx
```diff
@@ -0,0 +1,139 @@
...
+          teamId={pregameModel.teamId}
+          answerChoices={answerChoices}
+          score={score}
+          handleUpdateScore={handleUpdateScore}
+        />
+      );
+    case GameSessionState.PHASE_2_START:
+      return <StartPhase2 />;
+    case GameSessionState.FINAL_RESULTS:
+    default:
```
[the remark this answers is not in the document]
drewjhart: Totally agree. This routing was done in the `rejoin` PR

play/src/hooks/useFetchAndSubscribeGameSession.tsx
drewjhart: Moved.

play/src/lib/PlayModels.tsx
drewjhart: ugh, thanks. all string values removed here

play/src/pages/pregame/HowToPlay.tsx
drewjhart: nice, map statement has been added with <HowToPlaySlideXContent /> in an array

play/src/components/GameSessionSwitch.tsx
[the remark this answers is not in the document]
maniramezan: Yeah definitely, I think we should revisit on what are the next pain points are after separating containers and how we can help with reducing boilerplate and unnecessary null checks

play/src/hooks/useFetchAndSubscribeGameSession.tsx
drewjhart: I think the weird part here is that the .that the try/catch on fetching the gamesession and the try/catch on the subscribing gamesession are separate async functions (one using .then and one using async/await). So the nested try/catch won't actually propagate up to the enclosing one (I believe). 

This got me to dig more into the way errors are being thrown around the subscription though. Currently, there is some error catching but it doesn't throw explicit errors. Additionally, feeding in an invalid gamesession doesn't throw an error from the subscribe function.

The only thing the nested try/catch tries to do right now is catch errors from the initial subscription. The only way `play` impacts this is via the gameSession.id that it passes through (and that it receives directly from the fetch prior). I think disruptions between the fetch and subscribe in the gameSession.id will be rare and all of that logic is already being tested in `networking`. Therefore, I feel that the nested try/catch is only confusing the code here. I've replaced it with a check on the `fetchedGameSession.id` prior to the subscription as a final layer of error checking here.

drewjhart:
I'm going to merge this, Mani. An error checking PR will follow for the functions in GameSessionSwitch and maybe we can talk about the righton/networking model for IGameSession later in the week.

PR #650, opened 2023-06-07
[Central] - Create/edit questions flow - Bug fixes

Alexandra-Kushnirsky:
**Issue:**
This PR addresses the comments in git issue #621:   
To summarize, the create/edit questions flow had the following three bugs: 
(1) null values in questions produced invalid games, 
(2) Editing a correct answer’s steps added a progressively larger number of steps to the correct answer in that question, and 
(3) editing a question did not register as a change when updating a game (and thus “save game” button was disabled)

**Resolution:**
Regarding item 1: Central has been updated so that relevant feedback is provided to the user when a mandatory field is not populated (the game will not be created until all of these fields have been provided with user input). 
Regarding item 2: This issue was caused by a lack of parsing the instructions into an array before handling a mapping and saving of the updated steps in the correct answer. The solution was to parse the existing instructions when initializing the question. Furthermore, when question save is handled, only non-empty instructions will be saved to avoid UI clutter and unnecessary/redundant instructions being saved in the database.
Regarding item 3: Solution consisted of adding onChange event to the question form so that editing a question will disable the button. 

Relevant code:
- Item 1: handled in the function `handleSaveQuestion` in the file `CreateQuestion.jsx (lines 103-112)`
- Item 2: handled in the beginning of the file `CreateQuestion.jsx` in the function `QuestionForm on line 24` to parse instructions, and in the file `CreateQuestion.jsx` in the function `handleSaveQuestion on line 127` to filter out empty steps
- Item 3: handled in the file `GameMaker.jsx` in the `QuestionForm` component on `line 361`

**Preview:**

A bit difficult to capture all of the above fixes with a screenshot but below is a question with 3 steps being edited. After clicking the "add step" button, only one additional empty step displays rather than a much larger amount as before: 
Screenshot 2023-06-07 at 6 37 07 PM

(apologies for the lag) Here is a screen recording of each of the fixes above. We can see that the user cannot create a game until all necessary fields are populated, the user is able to edit the instructions for the correct answer without over production of steps, empty steps are deleted, and the user's changes to a questions registers as an update (enabling the "save game" button on the game maker screen:

https://github.com/rightoneducation/righton-app/assets/106450593/f4559403-3f25-484b-a52f-7f2716d10d69

@drewjhart , @maniramezan


central/src/components/CreateQuestion.jsx
```diff
@@ -97,12 +96,24 @@ export default function QuestionForm({ updateQuestion, question: initialState, g
       window.alert("Please enter a question");
       return;
     }
-
```
drewjhart, line 100: nice job making use of the existing window.alert functionality to catch more errors. This will help a lot when we go coordinate this with U/X

central/src/components/CreateQuestion.jsx
```diff
@@ -111,11 +122,9 @@ export default function QuestionForm({ updateQuestion, question: initialState, g
       window.alert("Please enter a cluster to save the game");
       return;
     }
-
     const questionToSend = { ...question }
     questionToSend.choices = JSON.stringify(questionToSend.choices)
-    questionToSend.instructions = JSON.stringify(questionToSend.instructions);
-
+    questionToSend.instructions = JSON.stringify(questionToSend.instructions.filter(step => step !== ""));
```
drewjhart, line 127: nice, that's awesome

drewjhart (approved):
looks good to me! Also, great write up on the PR description. That made it super easy to pull down the branch, test it and read the code. Let's touchbase on paired programming tomorrow to make sure the description wasn't too much of a pain to create and, if it was, where we can cut down detail.

central/src/components/CreateQuestion.jsx
```diff
@@ -97,12 +96,24 @@ export default function QuestionForm({ updateQuestion, question: initialState, g
       window.alert("Please enter a question");
       return;
     }
-
+    if (question.choices[0].text == null || question.choices[0].text === "") {
```
maniramezan, line 99: Since checking the string to `null` and empty string is happening in two places, probably better to create a method for it like:

```javascript
const isNullOrEmpty = (str) => {
    return !str || str.length // or str === null || str === ""
}
```

central/src/components/CreateQuestion.jsx
```diff
@@ -97,12 +96,24 @@ export default function QuestionForm({ updateQuestion, question: initialState, g
...
+      window.alert("Please enter a correct answer")
+      return;
+    }
+    if (question.instructions == null || question.instructions.filter(step => step !== "").length < 1) {
+      window.alert("Please provide at least one step for the correct answer explanation");
+      return;
+    }
+    for (let choiceI = 1; choiceI < (question.choices).length; choiceI++) {
```
maniramezan, line 107: It's usually better not to shorten, like `choiceI` vs. `choiceIndex` or just go with `index`. `index` can be shorten to `i` or `idx` as it's very common and used often, but when it's appendix, you can shorten it to `choiceIdx` for example but not `I` as latter is not common

central/src/components/CreateQuestion.jsx
```diff
@@ -97,12 +96,24 @@ export default function QuestionForm({ updateQuestion, question: initialState, g
...
+      if (question.choices[choiceI].text == null || question.choices[choiceI].text === "") {
+        window.alert("Please enter an answer for wrong answer " + choiceI);
```
maniramezan, line 109: nit: Maybe use string interpolation style for readability

```javascript
window.alert(`Please enter an answer for wrong answer  ${choiceI}`);
```

maniramezan (approved):
Great job on PR description and the video

Issue #652, opened 2023-06-07
[Play] - Leaderboard teams regenerate on each resize of screen. 


drewjhart:
On the final leaderboard screen, we don't currently store team avatar data on the back end so we are randomly generating other players' teams for this screen. When we resize the screen, these values are regenerated which is pretty confusing. To resolve this issue, we'll need to prevent the randomize code running on resize.

PR #656, opened 2023-06-09
[Play] - Leaderboard teams regenerate on each resize of screen.

josephinaim:
**Issue:**

Solves the issue of the team avatar values being regenerated and randomized every time the screen is resized. Github: https://github.com/rightoneducation/righton-app/issues/652

**Resolution:**

Originally the team avatars were being set at a random value which would regenerate each time the component was re-rendering, including when the screen would be resized. I took it outside of the component function and used a useMemo hook so that it is randomized only once and then stored and returned. It is only regenerated when the teams or teamId also changes.

- `Leaderboard.tsx` - this is where the team avatars values are being generated

**Preview:**

https://github.com/rightoneducation/righton-app/assets/79948102/8665d395-7224-41bb-9621-5b675fa51cc1

Video of the problem^^

https://github.com/rightoneducation/righton-app/assets/79948102/4978e09c-9690-4c4b-a24f-679925ed6c0a

Video of the change^^

@drewjhart


play/src/pages/finalresults/Leaderboard.tsx
```diff
@@ -65,6 +65,14 @@ export default function Leaderboard({
     }
   }, [containerRef.current?.clientHeight, subContainerHeight]); // updates whenever the container is resized
 
+  const randomAvatarNumbers = useMemo(
+    () =>
+      (teams ?? [])
```
maniramezan, line 70: You should be able to also just use `teams?.filter`

play/src/pages/finalresults/Leaderboard.tsx
```diff
@@ -65,6 +65,14 @@ export default function Leaderboard({
...
+        .filter((team) => team.id !== teamId)
```
maniramezan, line 71: better to use `find` as we're expecting only one team with `teamId`. `find` returns after finding the first team that matches the given criteria while `filter` continues to rest of the elements and returns all that match the criteria.

drewjhart: For some context here, we currently don't store the teamMember's avatar on the backend. So when we get to the leaderboard page, we just randomize the other teams avatar selections to display each of the other teams info.

So in the below image, everyone but Alex's color/avatar is randomly generated because we don't have access to that information:
image

I think we can probably update the backend to store this information but right now we're just following `mobile` for the implementation. 

The issue that Josephina is addressing here is that in the original code, I did the randomisation in the render method, so every time the screen resizes the colors change like crazy. The approach here, I think is to move that function outside of the render method and create an array of `teams.length -1` elements, each containing a random number to then map to in the render function. 

That being said, @josephinaim I am running into a runtime error with this. Because the array is length teams - 1, when we use the index in the render method, if the player is not in the last position, we end up trying to access a value in the randomAvatarNumbers array that exceeds its length. For example, if `teams.length = 6`, `randomAvatarNumbers.length = 5`. Assuming the player isn't in last place, we'll have a call where `index = 6` and `randomAvatarNumbers[6] = undefined`, giving us an error. 

Additionally, I like the initial idea of using useMemo here to stop the rerender (especially over a useEffect with an empty dependency array) but it does cause me a bit of pause because what we're wrapping the useMemo in isn't very expensive. I think the predominant use of that hook is performance optimisation  so there might be a way to do it with a more appropriate hook. I gave this a shot:

` const { current: randomAvatarNumbers } = useRef<number[]>(
    (teams ?? [])
      .map(() => Math.floor(Math.random() * 6))
  );`
  
 And it seemed to work. It's kind of dumb because it just produces an array of the same length as `teams` to avoid the `undefined` error (so there's one value in the array that is never used). If there's a way to get around not creating the additional random number that we don't need, I'm totally all ears but I don't think it's worth getting too fancy with tracking two indexes or whatever. 
  
 I took a look at this to validate my thoughts on `useMemo`: https://blog.logrocket.com/when-not-to-use-usememo-react-hook/. It's honestly probably a bit subjective and this person seems quite opinionated but I appreciate the idea of using the hooks for really specific use cases, particularly if there are some associated (albeit minor) costs. Let me know what you think though.

maniramezan: Thanks for the details, if it makes it easier, we can definitely store this in backend, can be a good pair programming practice to learn how to do this for anybody who's interested. Also, if any chances like this comes up that can help the codebase, let me know and we can work on making it easier by changing the backend and APIs.

drewjhart: That's a great idea, I'll reach out next week to set something up. @Alexandra-Kushnirsky or @josephinaim, would you be interested in joining that too?

And Josephina: I think we can keep working on this/merge it to fix this issue. Then, when the backend is updated, we can create another PR to handle the teamAvatar.

josephinaim: Sounds good to me, I would be very interested in joining in on the paired programming! For clarification, the adjustment I should make for now with this pr is to change the hook with:  
```const { current: randomAvatarNumbers } = useRef<number[]>( (teams ?? []) .map(() => Math.floor(Math.random() * 6)) );```

instead? Since this is a placeholder until we can have the team avatars stored in the backend.

Alexandra-Kushnirsky: Same here, would be great to have some deeper knowledge of how storing data works in the backend!

drewjhart: > Sounds good to me, I would be very interested in joining in on the paired programming! For clarification, the adjustment I should make for now with this pr is to change the hook with: `const { current: randomAvatarNumbers } = useRef<number[]>( (teams ?? []) .map(() => Math.floor(Math.random() * 6)) );`
> 
> instead? Since this is a placeholder until we can have the team avatars stored in the backend. @drewjhart

I'd treat this as a first step rather than a full solution. I think some comments would help, including a TODO about the backend and you could try to fix the fact that this produces one extra value in the array than we need. My code was more proof-of-concept tbh.

maniramezan: Maybe we can do a session this Friday or next to go over this. There'll be probably some new updates to amplify as well since we last updated it, hopefully no breaking changes

drewjhart:
Closed via: https://github.com/rightoneducation/righton-app/pull/683

PR #661, opened 2023-06-13
[Play] Timer resets on refresh  

Alexandra-Kushnirsky:
Will try to add as much detail as possible since I changed around quite a bit of the timer handling on the play side (hopefully this is not overkill). Also, I think for all intents and purposes, this PR explanation would make a bit more sense if the relevant code and resolution sections were together for readability purposes so I did that below just for this one: 

**Issue:**
In addition to addressing the timer in play restarting on refresh, I made some changes to the timer on play so that it is more consistent with the timer in host. 

**Resolution:**
First, to address timer inconsistencies, I tried to structure the timer in play similarly to the way it is handled in host by removing the recursive timer countdown function and instead using useEffect as is in host
In `Timer.tsx`, I removed `updateTimer` and instead added `useEffect` (now on `line 53-64`), which decrements variable `currentTime` by 1 second internally (every 1000 milliseconds) and is continued to be called until the timer has run out
In this function, we also handle the first issue of timer resetting on refresh by storing the current time in `localstorage` on `line 57`
I also removed const `milliseconds` from `Timer.tsx`. I think the reason we were using milliseconds was to make the animation for the timer bar more smooth but I ran into some issues with this so just kept it out for the time being. 
specifically: subtracting one millisecond from currentMillis every 1 millisecond was much slower than should be, maybe because there is a delay in between each calculation? 
Second, to ensure local storage would be re-initialized at the start of a phase, I added additional setting of localstorage outside of `Timer.tsx`:
First, in `GameInProgress.tsx`, I added a `useEffect` (`lines 105-114`) that is called when the game state changes. This handles the case that the game has begun (all phases including choose_correct_answer and on), in which case, localstorage should be updated to phase1time prior to re-entering phase1 (here when we are in phase_2_discuss) and to phase2time prior to entering or re-entering phase2 (here when we are in phase_1_discuss)
Then, in `PregameContainer.tsx`, we handle the case where the game has not yet begun (before first choose_correct_answer). Here, in the function `handleGameCodeClick` on `line 75`, we initialize localstorage to be phase1time to prepare for the first choose_correct_answer phase of the game

**A couple of issues:**
On refresh, the delay is not taken into account so the timer will resume from its time right before the refresh. I wasn’t entirely sure how to take this delay into account since I assume the delay time will vary to some extent.
It was not very ideal that the timer handling is dispersed throughout several files but thought it might make the most sense to keep the logic for timer contained in Timer.tsx as much as was possible. 
This is more of a bug report than specific to the code I changed but I also noticed that for the first question of each game in the first phase, timer on host is always slightly ahead of play (by roughly 3 seconds). Not sure if this is the case but maybe the countdown on host begins with the pregame countdown causing this?

I do realize this is a lot of changed code so if there are any suggestions to improve this solution, I would be happy to revise.

**Preview:**
Here is an image displaying the problem prior to the above changes. We can see that after refresh, the timer on play resets:
Screenshot 2023-06-12 at 8 53 03 PM

Below is a screen recording of the game after the above changes. We can see that the timer resumes to it's time prior to the refresh (though it is behind host because of the delay for refreshing the page), and that the time re-initializes as wanted before the start of each phase of the game. We can also see there is a slight (3 second) discrepancy between timer on host and play for the first phase of the first question:

https://github.com/rightoneducation/righton-app/assets/106450593/674f4170-e248-463f-8968-9386315ee33f

@maniramezan , @drewjhart


drewjhart:
Thanks for the call, Alex - that was super productive. Here is a breakdown of the tasks that came out of the discussion:

1. If you can `pull` from `dev` and `merge` into your PR to make sure you have the most recent copy of the code, that'd be great. I pushed some code yesterday (including that StorageKey). 

2. For the timer itself, I think we can use the `requestionAnimationFrame` variant to ensure smooth operation of the linear progress bar. We can update the `localStorage` only every second to reduce performance impacts and we can do that by using `useMemo`. I've attached the adjusted `Timer.tsx` file here (in txt for github):  [Timerb.txt](https://github.com/rightoneducation/righton-app/files/11737883/Timerb.txt)

3. For the `localStorage` itself  let's use that `StorageKey` value from `PlayModels` to store the `currentTimer` value we were talking about. You can take a look at `PregameContainer.tsx` for reference. 

4. We'll need to add a `currentTime` value to the local storage object and we can do that after ln 132 in `PregameContainer.tsx`

5. We can then grab that `currentTimer` value from local storage with something like `const currentTimer = localModel.currentTimer` after ln 29 in `GameInProgressContainer.tsx` and then pass it down to GameInPRogress.tsx as a prop (`currentTimer={currentTimer}`)

6. We can then adjust the check on ln 
165 of `GameInProgress.tsx` and add another condition where if `hasRejoined` is true than we use `currentTimer` instead of `phaseOneTime` or `phaseTwoTime.`

7. This should allow us to remove the `useEffect` between ln 109-118 in `GameInProgress.tsx` 

We talked through a lot so let me know if any of this isn't clear! Also, I think this should work but if you run into any snags and you get stuck or hit edge cases, let me know as well.

Alexandra-Kushnirsky:
> Thanks for the call, Alex - that was super productive. Here is a breakdown of the tasks that came out of the discussion:
> 
> 1. If you can `pull` from `dev` and `merge` into your PR to make sure you have the most recent copy of the code, that'd be great. I pushed some code yesterday (including that StorageKey).
> 2. For the timer itself, I think we can use the `requestionAnimationFrame` variant to ensure smooth operation of the linear progress bar. We can update the `localStorage` only every second to reduce performance impacts and we can do that by using `useMemo`. I've attached the adjusted `Timer.tsx` file here (in txt for github):  [Timerb.txt](https://github.com/rightoneducation/righton-app/files/11737883/Timerb.txt)
> 3. For the `localStorage` itself  let's use that `StorageKey` value from `PlayModels` to store the `currentTimer` value we were talking about. You can take a look at `PregameContainer.tsx` for reference.
> 4. We'll need to add a `currentTime` value to the local storage object and we can do that after ln 132 in `PregameContainer.tsx`
> 5. We can then grab that `currentTimer` value from local storage with something like `const currentTimer = localModel.currentTimer` after ln 29 in `GameInProgressContainer.tsx` and then pass it down to GameInPRogress.tsx as a prop (`currentTimer={currentTimer}`)
> 6. We can then adjust the check on ln
>    165 of `GameInProgress.tsx` and add another condition where if `hasRejoined` is true than we use `currentTimer` instead of `phaseOneTime` or `phaseTwoTime.`
> 7. This should allow us to remove the `useEffect` between ln 109-118 in `GameInProgress.tsx`
> 
> We talked through a lot so let me know if any of this isn't clear! Also, I think this should work but if you run into any snags and you get stuck or hit edge cases, let me know as well.

Thanks, Drew, this is super helpful!

host/src/components/HeaderGame.jsx
```diff
@@ -46,14 +46,14 @@ const useStyles = makeStyles(() => ({
 }));
```
drewjhart, line 46: let's revert any changes here so as to remove this file from the PR. We should only be including `play` files directly relating to the timer refresh issue

play/src/components/Timer.tsx
```diff
@@ -63,19 +70,26 @@ export default function Timer({
...
-  function getTimerString(currentTimeInput: number) {
-    let sec = 0;
-    let secStr = '00';
-    let min = 0;
-    if (currentTimeInput >= 0) {
-      min = Math.floor(currentTimeInput / 60);
-      sec = Math.ceil(currentTimeInput % 60);
-      if (sec === 60) sec = 0;
-      secStr = sec < 10 ? `0${sec}` : `${sec}`;
+  const timerString = useMemo(() => {
```
drewjhart, line 73: I messed up the implementation of `useMemo` originally here so I've pushed a commit that resolves it. Sorry for any confusion caused by that.

play/src/containers/GameInProgressContainer.tsx
```diff
@@ -26,6 +26,7 @@ export function GameInProgressContainer(props: GameInProgressContainerProps) {
   // retreives game data from react-router loader
   // if no game data, redirects to splashscreen
   const localModel = useLoaderData() as LocalModel;
+  const currentTimer = localModel.currentTimer;
```
drewjhart, line 29: this will flag a linting error about object destructuring.

play/src/pages/GameInProgress.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
```
drewjhart, line 1: I think we can remove the useEffect import here (linting should flag this)

drewjhart:
Really nice job on this @Alexandra-Kushnirsky, this was a tricky one. Code looks great and I pulled the branch and it seems to be working well. One thing: would you mind linting the code by running `yarn lint` on the branch and then resolving any errors and warnings that come up? I think it's minor stuff but it will just be pulling out any imports we aren't using, making use of object destructuring, removing any comment.logs etc etc.  Additionally, if you can run `yarn format` that will apply prettier as well (although that might be happening automatically). If you get a ton of files changed via `yarn format` just discard the changes of any files that aren't included in the original PR. 

The goal is to just lint and format the code every time before we merge it into `dev`. 

Great job again on this though!

play/src/containers/GameInProgressContainer.tsx
Alexandra-Kushnirsky: Hm didn't seem to get an error here after linting. Do we not want to restructure the object here and just pass in localModel.currentTimer where we use currentTimer in the file instead?

Alexandra-Kushnirsky:
@drewjhart Thank you for all your help! Just a quick question before I push again: is there any way to only yarn format specific files? After running, I get like 88 changed files for some reason :-(

play/src/containers/GameInProgressContainer.tsx
drewjhart: hmm, that's weird. I hope my linter settings are being carried over. I think for now we can just do `const { currentTimer} = localModel;` for consistency with the rest of the code

drewjhart:
> @drewjhart Thank you for all your help! Just a quick question before I push again: is there any way to only yarn format specific files? After running, I get like 88 changed files for some reason :-(

88 files yiiiiiikes. So I'll need to find a way to automate this (I think there's a way to set up linting/formatting through vscode to automatically do this). That's probably going to happen after the end of the month though. For now, we can use this really dumb command line solution:

```
// get list of files that differ from those on dev (this should match the files in the PR)
git diff --name-only dev...

// for each file in the list, run this to format via prettier
yarn prettier --write src/DIRECTORY/FILE.tsx
```

Note that the git command and the prettier command require different paths. Here's an example to illustrate:
```
git diff --name-only dev...
output: play/src/components/GameSessionSwitch.tsx
yarn prettier --write src/components/GameSessionSwitch.tsx
```

Sorry for the jank on this, I'll get a better solution for this once we get through the `play` stuff.

drewjhart (approved):

drewjhart:
Alright @Alexandra-Kushnirsky, I think the PR now just contains your changes and is linted and formatted. If you wouldn't mind reviewing the files to make sure I didn't overwrite anything on your end that would be great. I did a quick test but it'd be good to get your final sign off (this is totally a Monday thing!).

PR #670, opened 2023-06-20
[Play] - currentTimer null value produces error

drewjhart:
**Issue:**

In `GameInProgressContainer.tsx` ln 34 produces a null runtime error if the player navigates to play.rightoneducation.com/game, without joining a game first. The code is `const { currentTimer } = localModel;`

**Resolution:**

This is due to the fact that we're trying to destructure localModel.currentTimer when localModel is null. If we move `localModel.currentTimer` down to the render function on ln 81, this error is avoided because that function will only render when `localModel` is populated. Then, the `isNullOrUndefined` check on ln 45 won't be interrupted and the player will be routed back to `/` automatically. 

Relevant code:
- `GameInProgressContainer.tsx` - ln81 - move `localModel.currentTimer` to a location where `localModel` !== null


drewjhart:
Merging this now @Alexandra-Kushnirsky

Issue #672, opened 2023-06-20
[Play] - Design team comments for Congratulations screen


drewjhart:
The design team has finalized comments on the Congratulations screen [here](https://docs.google.com/document/d/1EyyDVn2hgrPhgxTok5Qf0AFVjNPh0PVeWxjKCWQtx2c/edit). From this, we can extract two `dev` tasks for this screen.

1. The Top 5 banner is displayed regardless of if a player is in the top 5 banner. 
Location in document: issue # 3 (on the bottom of Page 10). Resolution: We need to check if a player is in the top five and, if so, we display the "Great Job" banner that is linked in the above document. 

I think we are sorting the teams on the Leaderboard screen already. So we have the logic to check the team scores already. 

2. Text adjustment in Top 5 Banner.
Location in document: Issue # 7 (top of Page 13). Text in banner to be adjusted per comments in the document. 

Let me know if you have any questions!

drewjhart:
Hi @Alexandra-Kushnirsky and @josephinaim, I'm reassigning this to Alex to balance out the tasks and make sure you've still got something to work on Alex. I think the majority of comments we will be facing from the Design team will be on the Congrats and Leaderboard screen but hopefully we can divide up tasks to prevent any major merge conflicts. We can talk more about this on Thursday's paired programming.

PR #684, opened 2023-06-22
[Play] -  Congratulations screen

Alexandra-Kushnirsky:
**Issue:**
This PR addresses git issue #672:
(1) The Top 5 banner is displayed regardless of if a player is in the top 5
(2) 'Top 5' should be capitalized on 'Top 5' banner of congratulations screen

**Resolution:**
(1) Provide additional logic handling to pass into `Congrats.tsx` to determine if 'Top 5' banner should be displayed or 'Great Job' banner
(2) Adjust text item for top 5 banner in `translation.json`

Relevant code:
For item 1:
-Brought down logic handling of determining if the player is in the top five from previous `const leader` in `GameSessionSwitch.tsx` to `FinalResultsContainer.tsx` by removing prop `leader` from `FinalResultsContainer` in `FinalResultsContainer.tsx` and defining the following within that file:
-created function `isLeader` on `lines 34-45` which takes in the teams and teamId of this player as arguments and returns true if that player is in the top 5 and false otherwise. (I used a for loop instead of `.indexOf` since the complexity is O(5) vs O(n) using `.indexOf`. However, I do realize it’s not likely that a single game will have so many players that that would make a huge difference so if using `.indexOf` for readability purposes makes more sense, I can definitely change it)
-Then that function is used to determine what value is passed in as the `leader` prop of the `<Congrats>` component on `lines 61-65` 
For item 2:
-Capitalized 't' on `line 100` of `translation.tsx`

**Preview:**
This screenshot shows the corrected banner for players in the top 5 with capitalized t: 
Screenshot 2023-06-22 at 4 43 21 PM

For the below, I simulated a game with 6 players (with the player being run from local host in 6th place)
This is a screenrecording for the screen visible to players not in the top 5 which displays the 'Great job!' banner instead of the previous 'Top 5' banner:

https://github.com/rightoneducation/righton-app/assets/106450593/b10ccb11-2ccd-456f-a3d8-9d117ac37cea

@drewjhart @maniramezan

### DIFF
--- a/play/src/components/GameSessionSwitch.tsx
+++ b/play/src/components/GameSessionSwitch.tsx
@@ -35,7 +35,8 @@ export default function GameSessionSwitch({
   const { currentState } = gameSession;
   const currentQuestion =
     gameSession.questions[gameSession.currentQuestionIndex!]; // eslint-disable-line @typescript-eslint/no-non-null-assertion
-  const currentTeam = gameSession.teams!.find( // eslint-disable-line @typescript-eslint/no-non-null-assertion
+  const currentTeam = gameSession.teams!.find(
+    // eslint-disable-line @typescript-eslint/no-non-null-assertion
     (team) => team.id === localModel.teamId
   );
   // locally held score value for duration of gameSession, updates backend during each PHASE_X_RESULTS
--- a/play/src/components/GameSessionSwitch.tsx
+++ b/play/src/components/GameSessionSwitch.tsx
@@ -35,8 +35,7 @@ export default function GameSessionSwitch({
   const { currentState } = gameSession;
   const currentQuestion =
     gameSession.questions[gameSession.currentQuestionIndex!]; // eslint-disable-line @typescript-eslint/no-non-null-assertion
-  const currentTeam = gameSession.teams!.find(
-    // eslint-disable-line @typescript-eslint/no-non-null-assertion
+  const currentTeam = gameSession.teams!.find( // eslint-disable-line @typescript-eslint/no-non-null-assertion
     (team) => team.id === localModel.teamId
   );
   // locally held score value for duration of gameSession, updates backend during each PHASE_X_RESULTS
--- a/play/src/components/GameSessionSwitch.tsx
+++ b/play/src/components/GameSessionSwitch.tsx
@@ -40,7 +40,6 @@ export default function GameSessionSwitch({
   );
   // locally held score value for duration of gameSession, updates backend during each PHASE_X_RESULTS
   const score = currentTeam?.score ?? 0;
-  const leader = true;
   // this condition is used to display the pregamecountdown only on initial game start
   // this prevents a player from rejoining into the first screen and continually getting the pregame countdown
   // placed into a separate variable for readability in the switch statement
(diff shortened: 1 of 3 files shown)


play/src/components/GameSessionSwitch.tsx
```diff
@@ -35,12 +35,12 @@ export default function GameSessionSwitch({
   const { currentState } = gameSession;
   const currentQuestion =
     gameSession.questions[gameSession.currentQuestionIndex!]; // eslint-disable-line @typescript-eslint/no-non-null-assertion
-  const currentTeam = gameSession.teams!.find( // eslint-disable-line @typescript-eslint/no-non-null-assertion
+  const currentTeam = gameSession.teams!.find(
+    // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
drewjhart, line 39: do you mind bringing this // eslint-disable-line back to the original place? This will keep it from showing up in this PR

play/src/containers/FinalResultsContainer.tsx
```diff
@@ -21,14 +24,26 @@ export default function FinalResultsContainer({
...
   const [finalResultsState, setFinalResultsState] = useState(
     FinalResultsState.CONGRATS
   );
 
+  const isLeader = (inputTeams: ITeam[], inputTeamID: string) => {
+    inputTeams.sort((a, b) => b.score - a.score);
+    if (inputTeams.length <= 5) {
+      return true;
+    }
+    for (let idx = 0; idx < 5; idx += 1) {
```
drewjhart, line 39: I like the idea here of constraining what we are searching to to just the first 5 entires (rather than mapping through the entire array for instance). We could still make use of array methods here with something like this though:

`    if (inputTeams.slice(0, 5).find((team) => team.id === inputTeamID))
      return true;`
      
Feel like this might be more compact than the for loop

play/src/containers/FinalResultsContainer.tsx
```diff
@@ -46,7 +61,11 @@ export default function FinalResultsContainer({
         <Congrats
           score={score}
           selectedAvatar={selectedAvatar}
-          leader={leader}
+          leader={
+            !isNullOrUndefined(teams) && !isNullOrUndefined(teamId)
+              ? isLeader(teams, teamId)
```
drewjhart, line 66: we're going to be calling isLeader every time this renders. Can we use useRef like Josephina did in this [PR](https://github.com/rightoneducation/righton-app/pull/683/files) to prevent this from happening? we might want to move the `!isNullOrUndefined(teams) && !isNullOrUndefined(teamId)` into the function itself as well.

drewjhart (changes requested):
Thanks for getting on this so quickly, Alex. Some minor comments below and then should be good to go

play/src/containers/FinalResultsContainer.tsx
Alexandra-Kushnirsky: Yes, that's so much better

drewjhart (approved):

drewjhart:
This looks good, @Alexandra-Kushnirsky. Let's hold off on merging this until we get the backend issue resolved though. I'd like to do one final test before we push it to `dev`

Alexandra-Kushnirsky:
> This looks good, @Alexandra-Kushnirsky. Let's hold off on merging this until we get the backend issue resolved though. I'd like to do one final test before we push it to `dev`

Sounds good! Thanks for taking a look at this btw I know it's been a hectic day

drewjhart:
Did a quick test and looks good. Merging now

PR #681, opened 2023-06-22
[Play] - Unit Tests - Pregame Tests

drewjhart:
_Note: this is PR is one of a series of nested PRs that provide unit testing via Jest for `play`. They breakdown as such:
[[Play] - Unit Tests - Framework via Jest #679](https://github.com/rightoneducation/righton-app/pull/679)
->[[Play] - Unit Tests - Pregame Tests #681](https://github.com/rightoneducation/righton-app/pull/681)
->->[[Play] - Unit Tests - GameInProgress Test #685](https://github.com/rightoneducation/righton-app/pull/685)
->->->[[Play] - Unit Tests - Linting and Setup Revisions #693](https://github.com/rightoneducation/righton-app/pull/693)_

_Upon completion of review/comments: they will be merged per: 693->685->681->679_

**Issue:**
This PR provides testing for functions within the Pregame flow of the `play` app. Pregame here includes the Splash Screen to the Avatar Select Page and covers any code contained in `PregameContainer.tsx.` The functionality that we want to test is below:

- Valid text input on Enter Game Code screen.
- Valid text input on Enter Play Name screen. 
- Onclick on button of Select Avatar screen should fire addTeam callback
- Error should reveal Error Modal.

Text input is validated through `HelperFunctions.ts` so additional tests were also written for other functions in that file. These include:

- Verify if submitted answer is present on rejoin
- Validate local model on rejoin

**Resolution:**
Specific tests per file:

`EnterGameCode.test.tsx`:

1. Check that input text fields are rendering 
2. Check text validation via querying of DOM elements and mocking of input (test of validation function in HelperFunctions.test.tsx).
3. Check that button fires handleGameCodeClick

`EnterPlayerName.test.tsx`:
1. Check that input fields are rendering and error text is not rendering
2. Check that if text field is placeholder and submitted via button click, error text is rendered
3. Check that if names are valid and submitted via button click, error text is not rendered (test of validation function is conducted in HelperFunctions.test.tsx)

`SelectAvatar.test.tsx`:
1. Check that six avatars and button are rendered
2. Check that button click registers add team call back
3. Check that if error received, error modal is popped up

`HelperFunctions.test.tsx`:
1. Check that name is valid, isn't empty, isn't placeholder value
2. Check that entergamecode is valid, is 4 numbers, not strings, no placeholder, not empty
3. Check if an answer is submitted when a player rejoins. If state is in an answering phase and if answer has been added to gameSession object that aligns with teamID and phase, answer is submitted. 
[...]

PR #685, opened 2023-06-23
[Play] - Unit Tests - GameInProgress Test

drewjhart:
_Note: this is PR is one of a series of nested PRs that provide unit testing via Jest for `play`. They breakdown as such:
[[Play] - Unit Tests - Framework via Jest #679](https://github.com/rightoneducation/righton-app/pull/679)
->[[Play] - Unit Tests - Pregame Tests #681](https://github.com/rightoneducation/righton-app/pull/681)
->->[[Play] - Unit Tests - GameInProgress Test #685](https://github.com/rightoneducation/righton-app/pull/685)
->->->[[Play] - Unit Tests - Linting and Setup Revisions #693](https://github.com/rightoneducation/righton-app/pull/693)_

_Upon completion of review/comments: they will be merged per: 693->685->681->679_

**Issue:**
This PR provides testing for functions within the GameInProgress flow of the `play` app. GameInProgress here includes the Lobby through to the Final Results Leaderboard and covers any code contained in `GameInProgressContainer.tsx` and `FinalResultsContainer.tsx`. The functionality that we want to test is below:

- Validate that the correct Lobby/Game screen is rendered based on locally stored LocalModel data. 
- Validate that when the timer is stopped, a player is unable to answer a question.
- Validate that when the timer still has time remaining, a player is abler to answer a question. 
- Validate scoring by providing mock data to PhaseResults.tsx (this is in addition to any scoring validation in the networking layer)
- Validate that players are sorted correctly on the Leaderboard

**Resolution:**
Specific tests per file:

`GameInProgressContainer.test.tsx`:

1. Check that error modal pops if there's a connection error
2. Check that Rejoing text is rendered if gameSession is being rejoined into
3. Check that how to play Lobby is rendered  with loading text when gameSession is fetching for the first time
4. Check that pregame countdown is shown when subscription game state update is received

`GameInProgress.test.tsx`:
1. Check that if timer is finished, answer button is in disabled state
2. Check that if timer isn't finished, answer button is enabled (assumes player hasn't answered yet)

`PhaseResults.test.tsx`:
1. Check that in Phase 1 Result state, player correct, they receive 10 points to backend
2. Check that in Phase 1 Result state, player incorrect, they receive 0 points to backend
3. Check that in Phase 2 Result state, player picks most popular answer, they receive percent of players that picked answer
4. Check that in Phase 2 Result state, player picks most unpopular answer, they receive 0 points

`HelperFunctions.test.tsx`:
1. Check that players are sorted by score first and then alphabetically in the event of a tie

**Output:**
Screenshot 2023-06-28 at 3 21 35 PM
Screenshot 2023-06-28 at 3 21 49 PM
Screenshot 2023-06-28 at 3 21 57 PM

(6.5s with cache, 69s if yarn test --clearCache)


play/tests/unittests/mock/teams/sortedTeamsNoTies.json
maniramezan, line 1: For this and rest of mock data here, any reason we can't create these in code? That way we have compiler to help with any model changes, otherwise, if we modify the models, like what we did with `selectedAvatarIndex` we'd need to update all these

play/src/lib/HelperFunctions.tsx
```diff
@@ -129,4 +130,33 @@ export const fetchLocalData = () => {
...
+    );
+    const topTeamsSorted: ITeam[] = topTeamsUnsorted.sort((a, b) => {
+      if (a.score !== b.score) {
+        // sort by score descending
+        return b.score - a.score;
+      }
+      // sort alphabetically by name if scores are tied
+      return a.name.localeCompare(b.name);
+    });
+    return topTeamsSorted;
```
maniramezan, line 161: Although the above code works, but it's not efficient, it's iterating the teams multiple times. Being efficient for this part isn't that much important in our use-case as we have small number of teams, but if the number of teams increases significantly, modifying this to more efficient algorithm would be necessary. I'm going to use this as an opportunity to discuss some boring stuff about algorithms and runtime so feel free to ignore if not interested in them.

One thing that you might be interested in reading on is Insertion sort algorithm. Basically what it does is keeping the array sorted when a new items are added. This algorithm runtime is `O(n^2)` which makes it not efficient when the data is large, but for small numbers, IIRC arrays of size up to 20, this is an efficient algorithm to keep your array sorted while adding new items to it. This could be one approach to use here as our team size is less than 20. However, if we scale to support more teams, this wouldn't be the most efficient way of doing it.

You can take advantage of existing `sort`, `reverse` and the fact that in javascript `0` means `false` and anything else means `true`:

```typescript
inputTeams.sort((lhs, rhs) => {
    const scoreDiff = lhs.score - rhs.score
    if (scoreDiff === 0) {
        return lhs.name.localeCompare(rhs.name);
    }
    return scoreDiff
}).reverse().slice(0, 5)
```

one-liner version:

```typescript
const result = arr.sort((lhs, rhs) => {
     return lhs.score - rhs.score ?? lhs.name.localeCompare(rhs.name);
}).reverse().slice(0, 3)
```

play/src/lib/HelperFunctions.tsx
```diff
@@ -129,4 +130,33 @@ export const fetchLocalData = () => {
...
   return localModel;
+};
+
+/**
+ * sorts teams by score descending, then alphabetically by name
+ * only include teams with scores in the top five
+ * @param inputTeams - the teams to be sorted
+ * @returns - the sorted teams
+ */ 
+export const teamSorter = (inputTeams: ITeam[]) => {
```
maniramezan, line 141: This can be added as an extension to Array type in typescript, [this SO](https://stackoverflow.com/a/12803141/1450348) has some info, that'd be cleaner as then you can use this method by `inputTeams.sortByScore(5)` and also have the total number to be defined by the caller, like if in future you wanted to use this method, but only for top 2 teams, then you can

play/src/lib/HelperFunctions.tsx
maniramezan: cc @josephinaim @Alexandra-Kushnirsky in case if you're interested in the above details

play/tests/unittests/GameInProgress.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+      id: uuidv4(),
+      text: choice.text,
+      isCorrectAnswer: choice.isAnswer,
+      reason: choice.reason ?? '',
+  })) ?? [];
+};
+
+describe ('GameInProgress', () => {
+  // tests if timer is > 0, button on answer card is still enabled
+  it('Timer has time, isSubmitted === false', async () => {
```
maniramezan, line 75: Same comment on the other PR regarding making the description more English friendly for all tests in this PR and probably other PRs

play/tests/unittests/PhaseResults.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+      act(() => {
+        renderWithThemeRouterTranslation(gameSession, mockAnswerChoices);
+      });
+  
+      await waitFor(() => {
+        expect(apiClient.updateTeam).toHaveBeenCalled()});
+  
+      // tests that new score indicator has value of +10
+      expect(apiClient.updateTeam).toHaveBeenCalledWith({
+        id: gameSession.teams![0].id,
```
maniramezan, line 161: Same comment, in general, in tests, we should always confirm `teams` is not `null` and then use forced unwrap `!`

play/tests/unittests/PhaseResults.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+ReactModal.setAppElement('body');
+
+// Mock the ApiClient.updateTeam method
+jest.mock('@righton/networking', () => ({
+  ...jest.requireActual('@righton/networking'),
+  ApiClient: jest.fn().mockImplementation(() => ({
+    updateTeam: jest.fn(),
+  })),
+}));
+const apiClient = new ApiClient(Environment.Staging);
```
maniramezan, line 35: Can this be done in a separate file and then referenced in test files, we should only have one reference to `ApiClient(Environment.Staging)`. If we decide to switch environment or change the `ApiClient` to receive extra parameters, that would break all unit tests and requires going to all files to update them. We can probably use one of the global methods mentioned [here](https://jestjs.io/docs/api)

maniramezan (approved):

play/src/lib/HelperFunctions.tsx
drewjhart: Super cool explanation here and thanks for the code samples - those are so clean. For some context, this function is used to sort the Leaderboard at the end of a game. The requirements of that leaderboard are as follows:

1. Only show teams with a score in the top 5
2. If there are no ties, only 5 teams will be displayed.
3. If there are tied teams within this top 5, sort those tied teams alphabetically. More than 5 teams can be displayed. 

Your code is similar to what Josephina initially had for the sorting [here](https://github.com/rightoneducation/righton-app/pull/690/files) but unfortunately, we also need to ensure that we're including all possible teams within the top 5. 

Josephina and I were discussing this yesterday and this morning. Here's some of our thought process: https://github.com/rightoneducation/righton-app/pull/710. See my comment regarding the ordering of the sorts to cap the length of the search.

Basically, if we're starting with an unordered array, we need to sort the teams by score. Then, within that group of top five scores, we need to find the ties and sort each of those subsets alphabetically. Both Josephina and I were struggling with how to consolidate those two requirements into a single pass. What we can up with was to generate a set of unique scores, sort those, and then filter the existing array by that set. Then we could sort any ties within that limited set alphabetically. 

It would be nice if we could just count the number of ties and adjust the slice accordingly, like so:

```export const teamSorter = (inputTeams: ITeam[]) => {
  let ties = 0;
  return inputTeams.sort((lhs, rhs) => {
    const scoreDiff = lhs.score - rhs.score
    if (scoreDiff === 0) {
        ties =+ 1;
        return lhs.name.localeCompare(rhs.name);
    }
    return scoreDiff
  }).reverse().slice(0, 5 + ties);
};
```

But this blows up, because it is counting every tie (so if the top five has no ties and the bottom five are all tied, this will show the entire leaderboard). 

We were struggling with how to get around this issue and try to make it in a single pass. Do you have any other ideas to incorporate that final requirement?

Also, for my own education, could you explain how our original function is O(N^2)? I was thinking it was O(N) because we :

1. forEach - O(N)
2. sort - O(N)
3. filter.has - O(N) + O(1)
4. sort - O(N)

simplifies to O(N) (worst case: every team is in the top five, in some arrangement of ties).

I'm definitely still wrapping my head around Big O notation in practice so would you be able to steer me in where I'm going wrong on this? Thanks a ton for the comments @maniramezan!

play/tests/unittests/mock/teams/sortedTeamsNoTies.json
drewjhart: Great point, I hadn't thought about how difficult maintenance on these mock data sets would be. Reminds me of how we are pushing to enum a lot of our data fields so we don't hardcode strings. 

I remember you suggesting that our testing logic shouldn't provide any calls to the backend. So for creating this data by code, do we essentially need to mock the backend fucntions (createGameSession etc) that would be necessary here. I am assuming we would import the models from @righton/networking and then populate the objects accordingly? Is that the approach necessary?

play/src/lib/HelperFunctions.tsx
maniramezan: aww, sorry I missed that part, for it you can modify the above to something like:

```typescript
const newSortTeams = (teams: ITeam[]) => {
    const sortedTeams = teams.sort((lhs, rhs) => {
        return lhs.score - rhs.score ?? lhs.name.localeCompare(rhs.name);
   }).reverse()
   
   var lastScore = -1
   let totalTeams = 5
   let ret = Array(totalTeams)
   sortedTeams.every((team) => {
       if (team.score !== lastScore) {
           totalTeams -= 1
       }
       ret.push(team)
       return totalTeams !== 0
   })
   
   return ret
}
```

So the sort BigO is usually considered to be `nlogn` as mostly  they use quicksort. So generally, the proposed solution and the current solution both have the same BigO notation of `nlogn`, but memory footprint and also the actual performance would be different when you use a large set of data. 

So I started being curious about the actual performance comparison, so I created [this fiddle](https://jsfiddle.net/21xrd80y/2/). It looks like actually my proposed solution performed worse,  🤦‍♂️. I was pretty sure it'd be due to `reversed` method not being performant and doing some copy / iterating over the array work hence I modified the proposed solution to the following:

```typescript
const newNewSortTeams = (teams: ITeam[]) => {
    const sortedTeams = teams.sort((lhs, rhs) => {
        return lhs.score - rhs.score ?? lhs.name.localeCompare(rhs.name);
   })
   
   var lastScore = -1
   let totalTeams = 5
   let ret = Array(totalTeams)
   for (var i = sortedTeams.length - 1; i >=0 && totalTeams > 0; i--) {
       if (sortedTeams[i].score !== lastScore) {
           totalTeams -= 1
       }
       ret.push(sortedTeams[i])
   }
   
   return ret
}
```

And now, the numbers for running all three implementations for 10000000 are as follow:

```
sortTeams: 685.921875 ms
newSortTeams: 1706.89306640625 ms
newNewSortTeams: 618.531982421875 ms
```

As also mentioned before, the current implementation would suffice for current and any foreseeable version of the app. I was personally just curious on how this could be improved.

play/tests/unittests/mock/teams/sortedTeamsNoTies.json
maniramezan: Yeah, for these, we can use mocking version of APIClient and configure the return value. Something similar to `myMock.mockReturnValueOnce(10).mockReturnValueOnce('x').mockReturnValue(true);`  example [here](https://jestjs.io/docs/mock-functions)

drewjhart: Awesome, that clarifies it. I'll take a look and implement. Thanks for the example.

play/src/lib/HelperFunctions.tsx
drewjhart: This is so interesting. Wild that the plain for loop seems like the go to here. I'm signing off for the weekend but want to take a look at the fiddle and see what's going on. I'll probably just implement your revised though, as it seems both faster and easier to read. Also, thanks for the O(nlogn) explanation. totally makes sense with quicksort.

play/tests/unittests/mock/teams/sortedTeamsNoTies.json
drewjhart: I've gone through and rewrote the tests to mock ApiClient.ts and provide other mock functions similar to @righton/networking. This has removed the need for any .json mock data in the tests folders

play/src/lib/HelperFunctions.tsx
drewjhart: I decided to go with your for loop implementation here to save a few ms and because it seems cleaner (doesn't need sets etc).

play/src/lib/HelperFunctions.tsx
drewjhart: I might save the extension to the Array type for a future PR but have added a param for the number of teams to display so that it can be tuned per your suggestion

play/tests/unittests/GameInProgress.test.tsx
drewjhart: corrected

play/tests/unittests/PhaseResults.test.tsx
drewjhart: corrected

play/tests/unittests/PhaseResults.test.tsx
drewjhart: This has been done via our mock of ApiClient.ts

PR #686, opened 2023-06-23
Add selected avatar index to team model

maniramezan:
**Issue:**

Each team selects an avatar, but this data has always been local and in-memory so no way to go back to them in future to know which avatar was chosen by each team. These changes adds `selectedAvatarIndex` to backend to it stores by team's information.

**Resolution:**

Discussed [here](https://github.com/rightoneducation/righton-app/pull/656#discussion_r1224793798)

**Preview:**

No preview as it's only backend changes. Supposedly, this just shouldn't break any other codes.


debug-panel/package.json
```diff
@@ -24,8 +24,8 @@
     "web-vitals": "^3.3.1"
   },
   "scripts": {
-    "start": "yarn --cwd ../networking build && react-scripts start",
-    "build": "yarn --cwd ../networking build && react-scripts build",
+    "start": "yarn --cwd ../networking build && mkdir -p ./node_modules/@righton/networking && rsync -r ../networking/lib ../networking/src ../networking/package.json ./node_modules/@righton/networking && react-scripts start",
+    "build": "yarn --cwd ../networking build && mkdir -p ./node_modules/@righton/networking && rsync -r ../networking/lib ../networking/src ../networking/package.json ./node_modules/@righton/networking && react-scripts build",
```
maniramezan, line 28: @drewjhart This is what was missing, the changes in networking wasn't being updated under node_modules because we weren't copying it so it was using an older version. I've updated Play version to also matches here to both build and copy the latest networking.

drewjhart (approved):

drewjhart:
Hi Mani,
Thanks a ton for doing this and for taking the time to walk us through each of the steps. I've gone ahead and pulled built play off this branch and uploaded it to restore the functionality of adding a player. I'm going to leave this PR unmerged though until @Alexandra-Kushnirsky and @josephinaim have a chance to take a look at it (on Monday, of course) as I think it's fairly instructive. Thanks again for the work/teaching on this!

josephinaim (approved):

Alexandra-Kushnirsky (approved):

Issue #696, opened 2023-06-29
[Play] - Verify if we can remove the initial team sort on Leaderboard.tsx


drewjhart:
Issue:
From https://github.com/rightoneducation/righton-app/pull/695. In my review, I missed that we're still using the original `sortedTeams` logic as well as the new Set-based sorting. I'm pretty sure we can delete the original sortedTeams logic and just convert the set but it'd be good to verify this. 

Resolution:
This task is to verify that we can remove the original sorting logic (ln 39-53) on `Leaderboard.tsx`. If possible, delete that code and just use `teams`.

PR #695, opened 2023-06-29
[Play] -If there are no ties, only five players should be displayed on the leaderboard #691

josephinaim:
**Issue:**

The leaderboard currently shows up to six players on desktop, even when there are no ties.

Location in [Design Comments Document](https://docs.google.com/document/d/1EyyDVn2hgrPhgxTok5Qf0AFVjNPh0PVeWxjKCWQtx2c/edit): issue # 5 (on the bottom of Page 11).

Github item: https://github.com/rightoneducation/righton-app/issues/691

**Resolution:**

Leaderboard should show the teams with top five highest scores, no matter the number of teams or ties. I created a set that added every unique team score and converted that into an array. I sorted the array and took the top five highest scores and then I filtered through the sortedTeams array for any teams that have those scores.

Relevant code:
- Leaderboard.tsx

**Preview:**

Screen Shot 2023-06-28 at 8 13 18 PM

Current leaderboard^^^

Screen Shot 2023-06-28 at 7 46 47 PM

Updated leaderboard ^^^

PR #710, opened 2023-06-30
[Play] - Verify if we can remove the initial team sort on Leaderboard.tsx #696

josephinaim:
**Issue:**

Issue:
Issue on github item: "From https://github.com/rightoneducation/righton-app/pull/695. In my review, I missed that we're still using the original sortedTeams logic as well as the new Set-based sorting. I'm pretty sure we can delete the original sortedTeams logic and just convert the set but it'd be good to verify this."

Github item: https://github.com/rightoneducation/righton-app/issues/696

**Resolution:**

Proposed resolution on github item: "This task is to verify that we can remove the original sorting logic (ln 39-53) on Leaderboard.tsx. If possible, delete that code and just use teams."

I've changed the array that the set takes from sortedTeams to just teams. In my review of the code, I'm not sure how it would be possible to get rid of the original sorting logic (In 39-53). I know the main part we discussed for this item is that the sorting twice seems redundant. The first sorting is done on the original teams array, and includes ties being sorted alphabetically, repeats included.
The second sorting is done on the array of team scores, each one is unique, so no repeats. The first one has to be done separately as it considers all of the ties and puts them in order alphabetically. Because the second sorting has to be done on the array of team scores that have been converted from the set, the sorting is actually not being done on the same array of values. The two sorted arrays are necessary to obtain the array of all the teams that have the top five highest scores.

That being said, I could have full well missed a solution right under my nose and I'm open to suggestions.

Relevant code:
- `Leaderboard.tsx`

The leaderboard renders the same as the last PR:
Screen Shot 2023-06-29 at 10 48 26 PM

### DIFF
--- a/play/src/pages/finalresults/Leaderboard.tsx
+++ b/play/src/pages/finalresults/Leaderboard.tsx
@@ -37,7 +37,18 @@ export default function Leaderboard({
 }: LeaderboardProps) {
   const currentTeam = teams?.find((team) => team.id === teamId);
   const teamSorter = (inputTeams: ITeam[]) => {
-    inputTeams.sort((a, b) => {
+    // create a set to store unique scores with no repeats
+    const scoreSet = new Set<number>();
+    inputTeams.forEach((team) => scoreSet.add(team.score));
+    // convert the set to an array and sort it in descending order to retrieve only the top five highest scores
+    const sortedTopScores: Set<number> = new Set(Array.from(scoreSet)
+      .sort((a, b) => b - a)
+      .slice(0, 5));
+    // Filter through sortedTeams for all teams with the top five unique scores
+    const topTeamsUnsorted: ITeam[] = inputTeams.filter((team) =>
+      sortedTopScores.has(team.score)
+    );
+    const topTeamsSorted: ITeam[] = topTeamsUnsorted.sort((a, b) => {
       if (a.score !== b.score) {
         // sort by score descending
         return b.score - a.score;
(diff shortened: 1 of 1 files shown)


drewjhart:
Yeah, I get what you're saying with needing to apply a separate sorting logic for both the numeric and alphabetical components of the leaderboard. 

That being said, I think we can clean it up marginally. Currently (if we maintain `sortedTeams` on line 93), we're sorting the whole teams array by score/alphabet first. We then create a set based on all scores in that sortedTeams array (that still contains all teams) and sort that set again in descending order. So we're applying a sort of the same length twice. We can resolve this by first creating the set, sorting in descending order and slicing to top 5, and then applying our score/alphabet sort at the end. This means we're only sorting the entire team length once, and sorting the actual score/alphabet only for those teams with scores in the top 5.  Pretty marginal difference, but IDK why not. 

Also, I learned something from this. Apparently, `array.includes` is O(N) and `set.has` is O(1) ([reference](https://dev.to/arnaud/using-array-prototype-includes-vs-set-prototype-has-to-filter-arrays-41fg)). If we stick with `array.include` and nest it in the `.filter` we get O(N^2) but if we sub it out for `set.has`, we get O(N). Again, not a super big deal but might as well do it. 

I'd like to get this merged, and the changes are minor, so I just went ahead and pushed them. I'll merge the branch now.

josephinaim:
Oh, that is a great point. Thanks for giving a clear explanation, I think it makes sense for sure to make that change. @drewjhart

PR #778, opened 2023-07-31
Migrate to DynamoDB

maniramezan:
**Issue:**

Moving Central backend to Amplify / DynamoDB from RDS due to:

- RDS is more expensive than DynamoDB the way that we're using it
- DynamoDB backend is handled automatically through amplify so no more manually writing resolvers and keeping a backup
- Generated models already support filtering and pagination. This opens more room for working on central to add these features to homepage. This is both user-facing feature that helps to better use the app and also reduces the AWS costs as we read what we need
- Unifies backend across all products. This makes it easier to share code and strategy across projects

**Resolution:**

- Use the same amplify instance by just adding `Game` and `GameQuestion` models. Since `Question` already existed and used by `GameSession`, we use `GameQuestion`
- Instead of having one `APIClient` that does all the work, now it's broken than into one api client per model:
  - [BaseAPIClient](https://github.com/rightoneducation/righton-app/blob/b57b1d2b577f4b3a0db542bdaeeac0188d0cb58f/networking/src/APIClients/Impl/BaseAPIClient.ts) which includes all the helper methods needed to communicate with backend
  - _MODEL_APIClient: These models includes the methods needed for each model. For example, [GameSessionAPIClient](https://github.com/rightoneducation/righton-app/blob/b57b1d2b577f4b3a0db542bdaeeac0188d0cb58f/networking/src/APIClients/Impl/GameSessionAPIClient.ts) has all the methods related to  fetching, updating, and subscribing to `GameSession` that previously were in `APIClient.
  - This approach is more scalable and easier to manage as now it's more obvious which files needs to be updated when a backend change is needed
  - It also prevents us from getting to a place where `APIClient` is thousands of lines and hard to manage
  - This architectural change can be adopted gradually in `Player` project
- Parsers and AWS models have been moves into their own folders
- `AWSMobileApi` is now `GraphQLAPI`

**What's left:**
- Somehow writing tests for all api clients
- Update central with the model changes and fully test
- Deploy the changes to backend
- Make sure all api clients are using right models with interacting with amplify graphql api
  
**How to review:**
Review per commit instead of reviewing the whole PR. Adding new models has generated a lot of resolver files that you don't need to review at all. In addition, there're a lot of noise due to prettifier changes that we didn't have it run on all networking files so many files include both prettifier changes and code logic changes. This might make it hard to know what's new and what's just code formatting. However, going commit by commit should help you to know what that batch of files are intended to do.


drewjhart:
Hi Mani,
I've reviewed the files here, starting with the first commit and working up from there. This seems terrific and it's great to see us moving to DynamoDB. I know you've mentioned it previously so I appreciate you driving this forward.

That being said, some of these changes are honestly a bit over my head. I'm just finalizing availability with Alex and Josephina for our meeting on Friday. Once we get the time locked down, do you mind spending some time during that meeting to walk us through these changes? Be good to get a sense of implications, etc. Thanks again for this work!

maniramezan:
Hi Drew,

Yeah definitely, let me know what time works and we can meet and go over the changes. Some of it would actually be better to go over in person as I'm also interested in team's feedback on it.

drewjhart:
Closing this as our recent backend updates should cover this

Issue #709, opened 2023-06-29
[Play] - Confidence Meter - Functionality


drewjhart:
This is a broad task for the functional implementation of the Confidence meter, following the backend update.

This task will include:

1. Updating the backend via API call
2. Showing the 'sending response' text when api call is being sent (we do this on the `Lobby.tsx` screen) 
3. Ensuring that we don't have any race conditions on the backend (for example if someone just smashes the radio buttons, we don't want to send a dozen calls to the backend). We are doing this in `useFetchAndSubscribedGameSession.tsx` hook and it is based on [this](https://react.dev/learn/you-might-not-need-an-effect#fetching-data) in the React docs.

At completion of this tasks, the confidence meter feature should be completed.

PR #782, opened 2023-08-01
[Play] Confidence Meter API Call

Alexandra-Kushnirsky:
**Issue:**
This PR addresses subtask #709 of the new Confidence Meter feature: passing data between front and backend through API call and subscriptions

**Resolution:**
In networking: 

- added parameter `confidenceLevel` to the function `async updateTeamAnswer` on `lines 330, 335` of `ApiClient.ts` to be used in play

In play:
- `GameInProgress.tsx` - added support to be able to extract necessary fields to pass into above `updateTeamAnswer` function which is called on `line 223` in the function `handleSelectConfidence` (`lines 213 - 232`). In addition to backend updates, this function also updates the fields in the `selectConfidence` object accordingly to update the frontend. This object is defined on `lines 154-156` where helper function `checkForSelectedConfidenceOnRejoin` (defined in `HelperFunctions.tsx`) is called to repopulate fields accordingly if the player has rejoined the game. In addition to this, I've added/updated other constants in this file: (1) `displaySubmitted` on `lines 144-146` has been changed to account for existing subscription data in the case of rejoin (so that the confidence meter still appears if the player has submitted an answer when page is refreshed), (2) `currentAnswer` on `lines 147-149` is the player's answer to this question (if the player has submitted already) and is used as an argument to the helper function on `line 162` and to define the const `teamAnswerId` (3) `teamAnswerId` on `lines 150-152` is the id of the answer submitted by the player in this phase and is passed as an argument to `updateTeamAnswer` on `line 223` (4) To isolate frontend response between API errors caused on submit of answer vs confidence, removed const `isError` and added consts `isAnswerError` and `isConfidenceError` on `lines 73, 74`. These are used in `handleRetry` on `lines 190-203` to update the correct data when the error modal appears and player presses retry button and display the correct error modal on `lines 240-251`
- `HelperFunctions.tsx` - Added new function `checkForSelectedConfidenceOnRejoin` on `lines 96-131` to repopulate confidence data accordingly in the case that the player has refreshed the page
- `ErrorModal.tsx` - Added support for new confidence error type on `line 35`
- `PlayModels.tsx` - Added support for new confidence error type on `line 108`
- `translation.json (es + en)` - Added support for new text fields 
- `ChooseAnswer.tsx` - Added props to be passed down from `GameInProgress.tsx` to `ConfidenceMeter.tsx`
- `ConfidenceMeter.tsx` - updates parent's components as necessary and displays UI, responding to state changes coming from `GameInProgress.tsx`

**Preview:**
Here are some screen recordings of the new feature: 
The recording below demonstrates the following behavior: (1) Confidence meter appears in response to answer being submitted in phase 1 of gameplay (2) backend is updated with player's response (3) the player cannot submit more than one response within every 5 second interval (4) the player's selection is saved even after refreshing the page and (5) the 'sending response..' message appears when the API call is first made and becomes 'response sent' once API call is complete:

https://github.com/rightoneducation/righton-app/assets/106450593/8e3a75ee-2a77-4de4-9fdd-d81c2c880a2e

In the recording below I've created a dummy error in the code to illustrate what happens in the case of an API error when selecting confidence:

https://github.com/rightoneducation/righton-app/assets/106450593/fe4b7f9b-26c3-43b4-a3a7-98afa23804b4

In the recording below, we can see that if the player does not submit an answer within the phase, the confidence meter does not appear:

https://github.com/rightoneducation/righton-app/assets/106450593/ac595f9d-e9bf-4e72-844e-07d926311573

@drewjhart @maniramezan


play/src/components/ConfidenceMeterCard.tsx
```diff
@@ -0,0 +1,181 @@
...
+import { ConfidenceLevel, isNullOrUndefined } from '@righton/networking';
+import BodyCardStyled from '../lib/styledcomponents/BodyCardStyled';
+
+interface ConfidenceMeterCardProps {
+  selectedOption: number | null;
+  handleSelectOption: (index: number, confidence: ConfidenceLevel) => void;
+  isSelected: boolean;
+  isSmallDevice: boolean;
+  timeOfLastSelect: number | null;
+  setTimeOfLastSelect: (time: number | null) => void;
```
maniramezan, line 21: I might've missed it in the code, but doesn't look like `time` can be `null`

play/src/components/ConfidenceMeterCard.tsx
```diff
@@ -0,0 +1,181 @@
...
+          : t('gameinprogress.chooseanswer.sendingconfidenceresponse')}
+      </Typography>
+    </Box>
+  );
+
+  const handleRadioChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const currentTime = new Date().getTime() / 1000;
+    if (
+      isNullOrUndefined(timeOfLastSelect) ||
+      currentTime - timeOfLastSelect > 5
```
maniramezan, line 138: Usually when working with time, it'd be nice to comment on what intention is. This helps others checking the code to know what the interval is. One other way can be having the code documents itself:

```typescript
let currentEpochTimeInSeconds = new Date().getTime() / 1000;
```

play/src/lib/HelperFunctions.tsx
```diff
@@ -85,6 +86,50 @@ export const checkForSubmittedAnswerOnRejoin = (
...
+ * @returns - the index of the confidence the player has submitted, null if they haven't selected a confidence, boolean to track submission
+ */
+export const checkForSelectedConfidenceOnRejoin = (
+  hasRejoined: boolean,
+  currentAnswer: ITeamAnswer | null | undefined,
+  currentState: GameSessionState
+): {
+  selectedConfidenceIndex: number | null;
+  isSelected: boolean;
+  timeOfLastSelect: number | null;
```
maniramezan, line 103: hmm, is the purpose of having `null` is just because user hasn't selected anything? We can assign `0` if that's the case.

play/src/lib/HelperFunctions.tsx
```diff
@@ -85,6 +86,50 @@ export const checkForSubmittedAnswerOnRejoin = (
...
+      currentState === GameSessionState.CHOOSE_TRICKIEST_ANSWER)
+  ) {
+    if (!isNullOrUndefined(currentAnswer)) {
+      isSelected = currentAnswer.confidenceLevel !== ConfidenceLevel.NOT_RATED;
+      selectedConfidenceIndex = confidenceOptions.indexOf(
+        currentAnswer.confidenceLevel
+      );
+    }
+  }
+  return { selectedConfidenceIndex, isSelected, timeOfLastSelect };
```
maniramezan, line 130: We should still return the confidence enum itself instead of the index. Returning index is coupling the selected confidence with index of it in the array. If permutation changes or we remove / add new confidence, this can cause an error for index out of bound or incorrectly reports data

play/src/components/ConfidenceMeterCard.tsx
```diff
@@ -0,0 +1,181 @@
...
+  selectedOption,
+  handleSelectOption,
+  isSelected,
+  isSmallDevice,
+  timeOfLastSelect,
+  setTimeOfLastSelect,
+}: ConfidenceMeterCardProps) {
+  const theme = useTheme();
+  const { t } = useTranslation();
+  const confidenceOptionArray = [
```
maniramezan, line 34: nit: Might worth to create a type or interface and then use that to annotate the type of `confidenceOptionArray `, something like:

```typescript
interface IConfidenceOption {
    text: string
    value: ConfidenceLevel
}

let confidenceOptionArray: IConfidenceOption[] = [
    {
        text: "text",
        value: ConfidenceLevel.NOT_AT_ALL
    },
]
```

Why going over this amount of hassle? It's because then you've assigned a strong type to `confidenceOptionArray` variable. This way, if anybody tries to add an item that doesn't match the others or have a typo, typescript compiler would catch that error

play/src/pages/GameInProgress.tsx
```diff
@@ -144,41 +172,83 @@ export default function GameInProgress({
...
+      // after await, set isSelected to true again
+      setSelectConfidence((prev) => ({ ...prev, isSelected: false }));
+      await apiClient.updateTeamAnswer(teamAnswerId, true, confidence);
+      setSelectConfidence((prev) => ({
+        ...prev,
+        selectedConfidenceIndex: index,
+        isSelected: true,
+      }));
+    } catch {
+      setIsConfidenceError(true);
```
maniramezan, line 230: Should we set this to `false` at some point like when user retries selection?

play/src/pages/GameInProgress.tsx
```diff
@@ -136,6 +141,29 @@ export default function GameInProgress({
...
     return rejoinSubmittedAnswer;
   });
+  const [displaySubmitted, setDisplaySubmitted] = useState<boolean>(
+    !isNullOrUndefined(selectSubmitAnswer.selectedAnswerIndex)
+  );
+  const currentAnswer = teamAnswers?.find(
+    (answer) => answer?.questionId === currentQuestion.id
+  );
+  const [teamAnswerId, setTeamAnswerId] = useState<string>(
+    !isNullOrUndefined(currentAnswer) ? currentAnswer?.id : ''
```
maniramezan, line 151: Two notes:
1. You should be able to just use `currentAnswer.id` without `?`
2. In this case, you can use `??`: `currentAnswer?.id ?? ''` which should be both easier to write and more readable (latter part is debatable thought : D )

play/src/pages/gameinprogress/ChooseAnswer.tsx
```diff
@@ -90,31 +130,19 @@ export default function ChooseAnswer({
...
-                textAlign: 'center',
-                marginTop: `${theme.sizing.largePadding}px`,
-              }}
-            >
-              {t('gameinprogress.chooseanswer.answerthankyou2')}
-            </Typography>
-          </>
+          <Typography
+            sx={{
+              fontWeight: 700,
```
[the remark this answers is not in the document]
drewjhart: We've run out of the predescribed variants for Typography values ([here](https://mui.com/material-ui/customization/typography/#adding-amp-disabling-variants)) and will have to extend our theme. You're right though, we should try to link this up to the theme. @Alexandra-Kushnirsky, this is outside of the scope of your task so I'll add it to my todo list via https://github.com/rightoneducation/righton-app/issues/783

play/src/lib/HelperFunctions.tsx
drewjhart: Alex raised this with me in a meeting so probably poor advice on my end. The annoying thing is that mui radio buttons take a `value` prop that has to be a string, so we're stuck doing some sort of conversion. The initial idea was just to do this conversion as close to the component as possible (immediately before making use of the radio button component) but thinking about it, we could probably just stringify our enum and pass it to the radio button component. I think this would eliminate any discrepancy between the enum we use and the value we pass to the radio button. In other words, we would always be programmatically referencing the enum state so if it changes, the component will catch that change.

play/src/pages/gameinprogress/ChooseAnswer.tsx
```diff
@@ -69,6 +81,34 @@ export default function ChooseAnswer({
     </>
   );
 
+  const onSubmitDisplay =
+    currentState === GameSessionState.CHOOSE_CORRECT_ANSWER ? (
+      <Fade in={displaySubmitted} timeout={500}>
```
drewjhart, line 86: niiiice, this looks great in the video

play/src/pages/gameinprogress/ChooseAnswer.tsx
```diff
@@ -69,6 +81,34 @@ export default function ChooseAnswer({
     </>
   );
 
+  const onSubmitDisplay =
+    currentState === GameSessionState.CHOOSE_CORRECT_ANSWER ? (
```
drewjhart, line 85: Has the design team confirmed that this component can only appear in phase one of the game?

play/src/pages/gameinprogress/ChooseAnswer.tsx
```diff
@@ -90,31 +130,19 @@ export default function ChooseAnswer({
           selectedAnswer={selectedAnswer}
           handleSelectAnswer={handleSelectAnswer}
         />
+        {displaySubmitted ? onSubmitDisplay : null}
```
drewjhart, line 133: I think that this behaviour works for the in-progress branch of this overall Confidence feature. However, I think we'll need to coordinate with the Design team to see how they are planning to enable or disable this feature (I assume on the host side at some point). We can talk through this further on paired programming tomorrow

play/src/lib/HelperFunctions.tsx
Alexandra-Kushnirsky: Yes! That makes a lot more sense than representing time as null. I'll update all instances of timeOfLastSelect to address the above comments: (1) have the variable storing the `currentTime` in the `handleRadioChange` function in `ConfidenceMeter.tsx` document itself by renaming the var `currentEpochTimeInSeconds` and (2) change all typing of `timeOfLastSelect` from `number | null` to just `number`, giving it the value of 0 before the user interacts with the radio buttons

drewjhart (commented):
Following the comments above, the code looks good Alex. Nice to see this feature coming together in a very real way. One admin question, how does this PR work with this one: https://github.com/rightoneducation/righton-app/pull/712. Should we be merging this one into 712 or do we need to close 712 out and replace it with this one?

play/src/pages/GameInProgress.tsx
Alexandra-Kushnirsky: So currently, this is handled in the handleRetry function. The flow as of right now when a confidence-related API error occurs looks something like this: (1) `isConfidenceError` is set to true as seen above (2) `<ErrorModal>` appears as triggered by `isModalOpen={isConfidenceError}` on `line 247` (3) When the user clicks on the retry button, `handleRetry`, defined on `lines 190-203`, is called and when the condition `if (isConfidenceError)` is met, `setIsConfidenceError(false)` is called from `line 196`, setting the confidence error to false when user retries selection

Alexandra-Kushnirsky:
> Following the comments above, the code looks good Alex. Nice to see this feature coming together in a very real way. One admin question, how does this PR work with this one: #712. Should we be merging this one into 712 or do we need to close 712 out and replace it with this one?
@drewjhart Thanks for the feedback! and ahh I see what your saying. I can go ahead and close out #712 since I started this branch working off of the changes in #712. Should have just made a separate branch to merge but yes #712 is redundant as of now

play/src/pages/gameinprogress/ChooseAnswer.tsx
Alexandra-Kushnirsky: Yep, Yong Lin specified the card should only display on phase 1

play/src/pages/gameinprogress/ChooseAnswer.tsx
Alexandra-Kushnirsky: Sounds good!

play/src/pages/gameinprogress/ChooseAnswer.tsx
drewjhart: Perfect, thanks for confirming

play/src/lib/HelperFunctions.tsx
maniramezan: Hmm, if the conversion is needed to string, why not add it to the enum itself. Enums can have values in typescript, so you can assign string or int values to them. That's a nice feature that can be useful in these cases

drewjhart:
> > Following the comments above, the code looks good Alex. Nice to see this feature coming together in a very real way. One admin question, how does this PR work with this one: #712. Should we be merging this one into 712 or do we need to close 712 out and replace it with this one?
> > @drewjhart Thanks for the feedback! and ahh I see what your saying. I can go ahead and close out #712 since I started this branch working off of the changes in #712. Should have just made a separate branch to merge but yes #712 is redundant as of now

Nice, sounds good. I think we can then branch of this branch to create the host features and then nest the PRs. Then we'll be able to extend branches and PRs for the entire feature and merge the thing in when it's all done. We can cover this tomorrow as well

play/src/pages/GameInProgress.tsx
maniramezan: aww.nice, thanks for confirmation and the details : )

play/src/pages/GameInProgress.tsx
Alexandra-Kushnirsky: hmm `??: currentAnswer?.id ?? ''` seems to cause 'expression expected' issue so I've just removed the ? after currentAnswer.id and have it as this for now: `!isNullOrUndefined(currentAnswer) ? currentAnswer.id : ''`

play/src/lib/HelperFunctions.tsx
Alexandra-Kushnirsky: ahh ok I'll look into this right now but that's much more readable. I was sort of struggling to find a clean way of dealing with the conversions but the above is great advice, will take another look with all of that in mind!

Alexandra-Kushnirsky: @maniramezan I just had a quick question: my understanding is that the enum would be rewritten as follows: 
`enum ConfidenceLevel {
  NOT_RATED = "NOT_RATED"
  NOT_AT_ALL = "NOT_AT_ALL"
  KINDA = "KINDA"
  QUITE = "QUITE"
  VERY = "VERY"
  TOTALLY = "TOTALLY"
}` but would this change be in networking (meaning the backend would have to be updated again) or would this enum simply just be re-written somewhere in play and when the enum is referenced on the frontend, the string conversion is implied?

play/src/pages/GameInProgress.tsx
maniramezan: Great thanks! Interesting that `??` didn't work, I kinda expected  instead of `!isNullOrUndefined(currentAnswer) ? currentAnswer?.id : ''` you can just use `currentAnswer?.id ?? ''`

play/src/lib/HelperFunctions.tsx
maniramezan: @Alexandra-Kushnirsky hmm interesting, I forgot this enum is in graphql models and we haven't mapped it to a local enum type, right? If yes, then looks like the AWS generated model already has the string values for you, [ConfidenceLevel](https://github.com/rightoneducation/righton-app/blob/02630b0df5eed09ef1700c99fbffb81554d265a4/networking/src/AWSMobileApi.ts#L212). You can also go the re-written path. Latter gives us the opportunity to decouple frontend and backend, but also not necessary at the moment

Alexandra-Kushnirsky: Ok I see, thank you!

maniramezan (approved):
Thanks for detailed explanations and updates, nice to see this all coming together ❤️

Alexandra-Kushnirsky:
> Thanks for detailed explanations and updates, nice to see this all coming together ❤️

No, thank you! @maniramezan @drewjhart Really appreciate all of this feedback from everyone, it's been super helpful!

PR #791, opened 2023-08-22
[Backend] - Adds isHintEnabled and isConfidenceEnabled to Question

drewjhart:
**Issue:**

The new [Hint](https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?type=design&node-id=15530-192339&mode=dev) and [Confidence](https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?type=design&node-id=15173-144544&mode=dev) features are designed to be turned on and off on a question-by-question basis. Therefore, we need a way to control the enable/disable of these features at the `Question` level. 

I've added comments to files with relevant code. 

**Resolution:**

We have added the following fields to `Questions`:

```
isHintEnabled: boolean
isConfidenceEnabled: boolean
```

**Pending Discussion:**

@maniramezan I've also included the type updates (`IQuestion`, `ApiClient.ts`) in this pr (I haven't `amplify push`'d this yet).  Could you help me with something though?

This update contains only code for `networking` impacting `host` and `play`. One question though: how are the questions on this side (in DynamoDB, instead of RDS) created? I see that on `CreateGameSession` in `APIClient` hits an endpoint to generate the `GameSession`. Meanwhile, on the `central` side there are two API functions `CreateQuestion` and `CreateGameQuestion`. I think I need to set a default condition (similar to how we did it with the ConfidenceLevel parameter in Alex's last update) but am unsure where I can add that. Ideally, I'd like to avoid the `central` side as I'm trying to add these fields only to `GameSession` `Questions`.

### DIFF
--- a/networking/.graphqlconfig.yml
+++ b/networking/.graphqlconfig.yml
@@ -1,6 +1,6 @@
 projects:
   mobile:
-    schemaPath: amplify/backend/api/mobile/build/schema.graphql
+    schemaPath: src/graphql/schema.json
     includes:
       - src/graphql/**/*.ts
     excludes:


networking/src/ApiClient.ts
```diff
@@ -483,6 +483,8 @@ type AWSQuestion = {
     grade?: string | null
     gameSessionId: string
     order: number
+    isHintEnabled: boolean
```
drewjhart, line 486: fields added to AWSQuestion

networking/src/Models/IQuestion.ts
```diff
@@ -10,6 +10,8 @@ export interface IQuestion {
     grade?: string | null
     gameSessionId: string
     order: number
+    isHintEnabled: boolean
+    isConfidenceEnabled: boolean
 }
```
drewjhart, line 15: Fields added to models

networking/amplify/backend/api/mobile/schema.graphql
```diff
@@ -47,6 +47,8 @@ type Question @model(subscriptions: null, timestamps: null) {
   domain: String
   grade: String
   order: Int!
+  isHintEnabled: Boolean! @default(value: "false")
+  isConfidenceEnabled: Boolean! @default(value: "false")
   gameSessionId: ID! @index(name: "byGameSession", sortKeyFields: ["id"])
```
drewjhart, line 52: Fields added prior to `amplify gql-compile` and `amplify codegen`

maniramezan:
These questions are currently created through Lambda, moving toward integrating central to Dynamodb as well could help with even deleting lambda and just use normal GameSession create method from graphql. I can update the lambda with these changes when it's pushed to amplify.

drewjhart:
> These questions are currently created through Lambda, moving toward integrating central to Dynamodb as well could help with even deleting lambda and just use normal GameSession create method from graphql. I can update the lambda with these changes when it's pushed to amplify.

Hi @maniramezan, ah that makes a lot of sense. Thanks for the explanation. I just pushed to amplify now. Whenever you get a chance, would you be able to update the lambda? Thanks a ton for the help with this.

PR #807, opened 2023-09-26
[Backend] - Preliminary Backend Updates for Short Answer Response

drewjhart:
**Issue:**

We are currently implementing the ability for short answer responses into `play` and `host`. The short answer response component is implemented as a [Quill.js ](https://quilljs.com/) pad that can handle plain text and katex formulae entry. Additionally, it has the potential to handle Rich Text, which we will implement in the future. This allows students to enter a variety of text, on multiple lines, and submit it for entry.

As a result, the answers being submitted will be more complex than a simple radio button selection and we need to be able to store additional information in the TeamAnswer object. Currently, we only have a `text: string` field. I am proposing that we use `answerContents: AWSJSON` to store a JSON blob on the backend that will actually be an object that reflects the content of the quill notepad (with some cleaning for brevity and security). 

To speed in development, I'd like to be able to start using these backend fields as I hook up `play` and `host`, however, I don't want to break the current build. Therefore, I'm pushing a preliminary version of these changes that maintains the existing `text` field and defaults the `answerContents` to an empty string. This will prevent breaking the app but allow for development of `answerContents` on a separate branch.

**Resolution:**

A description of the resolution of the issue. Again, links to discussions are great. 

Relevant code:
- `schema.graphql` has been updated with the following: 

```
type Question @model(subscriptions: null, timestamps: null) {
  id: Int! @primaryKey(sortKeyFields: ["order", "gameSessionId"])
  text: String!
  choices: AWSJSON
  imageUrl: String
  instructions: AWSJSON
  standard: String
  cluster: String
  domain: String
  grade: String
  order: Int!
  isHintEnabled: Boolean! @default(value: "false")
  isConfidenceEnabled: Boolean! @default(value: "false")
  isShortAnswerEnabled: Boolean! @default(value: "false") // <-- new, false by default
  gameSessionId: ID! @index(name: "byGameSession", sortKeyFields: ["id"])
}

type TeamAnswer @model {
  id: ID!
  questionId: Int!
  isChosen: Boolean!
  text: String! // <-- will be removed in the future
  answerContents: AWSJSON! // <--- new
  isTrickAnswer: Boolean!
  confidenceLevel: ConfidenceLevel! @default(value: "NOT_RATED")
}
```
- `addTeamAnswer` in `ApiClient.ts` has been updated as per below:

```
 async addTeamAnswer(
        teamMemberId: string,
        questionId: number,
        text: string,
        isChosen: boolean = false,
        isTrickAnswer: boolean = false
    ): Promise<ITeamAnswer> {
        const tempAnswerContent = ''; // this is to prevent breaking the current deployment of play/host and will be removed in subsequent PRs
        const input: CreateTeamAnswerInput = {
            questionId,
            isChosen,
            isTrickAnswer,
            text, // leaving this in to prevent breaking current build, will be removed when answerContents is finalized
            answerContents: tempAnswerContent, 
            teamMemberAnswersId: teamMemberId,
            confidenceLevel: ConfidenceLevel.NOT_RATED
        }
        const variables: CreateTeamAnswerMutationVariables = { input }
        const answer = await this.callGraphQL<CreateTeamAnswerMutation>(
            createTeamAnswer,
            variables
        )
        if (
            isNullOrUndefined(answer.data) ||
            isNullOrUndefined(answer.data.createTeamAnswer)
        ) {
            throw new Error(`Failed to create team answer`)
        }
        return answer.data.createTeamAnswer as ITeamAnswer
    }
    ```

### CODE REVIEW

File: networking/src/ApiClient.ts [hunk 1]
```diff
@@ -921,6 +926,7 @@ class TeamAnswerParser {
             isChosen,
             isTrickAnswer,
             text,
+            answerContents,
             createdAt,
             updatedAt,
             teamMemberAnswersId,
```
maniramezan, line 932: At this point, we should cast this to a strongly-typed and parse the json file

maniramezan (approved):

play/src/mock/MockGameSession.json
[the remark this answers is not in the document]
drewjhart: Thanks, that's a good idea. I've created this issue to track it: https://github.com/rightoneducation/righton-app/issues/808. I think this will be in a separate PR, as it will require updates to our stories in `host` and `play` as well.

networking/src/ApiClient.ts
drewjhart: Yup, sounds good. I will create the parsing in the follow up PR when I start working with the actual answer content.

PR #838, opened 2023-10-17
[Networking] - Test-* environment set up

drewjhart:
**Issue:**

As we begin to make backend redesign changes, it has become more important to have multiple environments set up on AWS so that we can test changes while ensuring that a stable version is still available for the public. As such, we have created a stable `main` environment and a `test` environment for development. We are following: https://docs.aws.amazon.com/amplify/latest/userguide/team-workflows-with-amplify-cli-backend-environments.html for the main structure of this.

This PR adds the `testing` environment to networking. The testing environment is accessible via:

test-central.rightoneducation.com
test-host.rightoneducation.com
test-play.rightoneducation.com


drewjhart:
@maniramezan I don't think a code review is super necessary on this but just an FYI that the test- environment is good to go and the Lambda has been adjusted to work with "staging" and "testing" stages.

PR #856, opened 2023-10-30
[Play/Host] - Short Answer Response - Bug Fixes

drewjhart:
**Issue:**
This PR fixes some miscellaneous bugs based, for example if a student enter a plain number as an expression, how phase two answers are displayed after short answers in phase one, and preventing swiping when the quill pad is focused. 

_Note: this PR is a part of a multi-part PR for the Short Answer Responses feature. The full PR structure is below:
[[Play/Networking] - Short Answer Response Base #844](https://github.com/rightoneducation/righton-app/pull/844)
->[[Host] - Short Answer Response Base #845](https://github.com/rightoneducation/righton-app/pull/845)
->->[[Networking/Host] - Short Answer Response Updates to Backend #848](https://github.com/rightoneducation/righton-app/pull/848)
->->->[[Play] - Short Answer Response Scoring #849](https://github.com/rightoneducation/righton-app/pull/849)
->->->->[[Play/Host] - Game Flow #850](https://github.com/rightoneducation/righton-app/pull/850)
->->->->->[[Host] - Featured Mistakes #854](https://github.com/rightoneducation/righton-app/pull/854)
->->->->->->[[Play/Host] - Short Answer Formatting/Linting #855](https://github.com/rightoneducation/righton-app/pull/855)
->->->->->->->[[Play/Host] - Short Answer Response - Bug Fixes #856](https://github.com/rightoneducation/righton-app/pull/856)
->->->->->->->->[[Play/Host] - Short Answer - Helper Function Comments #857](https://github.com/rightoneducation/righton-app/pull/857)
->->->->->->->->->[[Host/Play/Networking] - Short Answer Response - Adjusts Answer Handling #858](https://github.com/rightoneducation/righton-app/pull/858)_

PR #857, opened 2023-10-30
[Play/Host] - Short Answer - Helper Function Comments

drewjhart:
**Issue:**
This PR adds some comments to the helper functions on `play` and `host` and adds references to the new documentation in the wiki, here: 

-[Short Answer Responses ‐ Overview](https://github.com/rightoneducation/righton-app/wiki/Short-Answer-Responses-%E2%80%90-Overview)
-[Short Answer Responses ‐ Answer Normalization]( https://github.com/rightoneducation/righton-app/wiki/Short-Answer-Responses-%E2%80%90-Answer-Normalization)
-[Short Answer Response ‐ Equality Checks](https://github.com/rightoneducation/righton-app/wiki/Short-Answer-Response-%E2%80%90-Equality-Checks)

_Note: this PR is a part of a multi-part PR for the Short Answer Responses feature. The full PR structure is below:
[[Play/Networking] - Short Answer Response Base #844](https://github.com/rightoneducation/righton-app/pull/844)
->[[Host] - Short Answer Response Base #845](https://github.com/rightoneducation/righton-app/pull/845)
->->[[Networking/Host] - Short Answer Response Updates to Backend #848](https://github.com/rightoneducation/righton-app/pull/848)
->->->[[Play] - Short Answer Response Scoring #849](https://github.com/rightoneducation/righton-app/pull/849)
->->->->[[Play/Host] - Game Flow #850](https://github.com/rightoneducation/righton-app/pull/850)
->->->->->[[Host] - Featured Mistakes #854](https://github.com/rightoneducation/righton-app/pull/854)
->->->->->->[[Play/Host] - Short Answer Formatting/Linting #855](https://github.com/rightoneducation/righton-app/pull/855)
->->->->->->->[[Play/Host] - Short Answer Response - Bug Fixes #856](https://github.com/rightoneducation/righton-app/pull/856)
->->->->->->->->[[Play/Host] - Short Answer - Helper Function Comments #857](https://github.com/rightoneducation/righton-app/pull/857)
->->->->->->->->->[[Host/Play/Networking] - Short Answer Response - Adjusts Answer Handling #858](https://github.com/rightoneducation/righton-app/pull/858)_

PR #854, opened 2023-10-26
[Host] - Featured Mistakes

drewjhart:
**Issue:**
This adds the Featured Mistakes component to `host` allowing teachers to select the answers that they will send to phase two so that students can pick them as popular mistakes. 

The Figma for the component can be found [here](https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?type=design&node-id=15530-192339&mode=design&t=pbEAbi4DOjD8ps1F-4)and it looks as follows:
Screenshot 2023-10-30 at 5 13 58 PM

**Preview**
You can see a recorded video of the feature below:

_Note: this PR is a part of a multi-part PR for the Short Answer Responses feature. The full PR structure is below:
[[Play/Networking] - Short Answer Response Base #844](https://github.com/rightoneducation/righton-app/pull/844)
->[[Host] - Short Answer Response Base #845](https://github.com/rightoneducation/righton-app/pull/845)
->->[[Networking/Host] - Short Answer Response Updates to Backend #848](https://github.com/rightoneducation/righton-app/pull/848)
->->->[[Play] - Short Answer Response Scoring #849](https://github.com/rightoneducation/righton-app/pull/849)
->->->->[[Play/Host] - Game Flow #850](https://github.com/rightoneducation/righton-app/pull/850)
->->->->->[[Host] - Featured Mistakes #854](https://github.com/rightoneducation/righton-app/pull/854)
->->->->->->[[Play/Host] - Short Answer Formatting/Linting #855](https://github.com/rightoneducation/righton-app/pull/855)
->->->->->->->[[Play/Host] - Short Answer Response - Bug Fixes #856](https://github.com/rightoneducation/righton-app/pull/856)
->->->->->->->->[[Play/Host] - Short Answer - Helper Function Comments #857](https://github.com/rightoneducation/righton-app/pull/857)
->->->->->->->->->[[Host/Play/Networking] - Short Answer Response - Adjusts Answer Handling #858](https://github.com/rightoneducation/righton-app/pull/858)_

PR #849, opened 2023-10-25
[Play] - Short Answer Response Scoring

drewjhart:
**Issues**
Adjusts the scoring to handle Short Answer Responses. The main difference here is that in Phase 1, if short answer response mode is selected, we use the IResponse in IQuestion to check for correctness instead of the IChoice. This means we only have to do the expensive equality checks as we receive them via CreateTeamAnswer subscription, instead of whenever we need to check correctness. 

_Note: this PR is a part of a multi-part PR for the Short Answer Responses feature. The full PR structure is below:
[[Play/Networking] - Short Answer Response Base #844](https://github.com/rightoneducation/righton-app/pull/844)
->[[Host] - Short Answer Response Base #845](https://github.com/rightoneducation/righton-app/pull/845)
->->[[Networking/Host] - Short Answer Response Updates to Backend #848](https://github.com/rightoneducation/righton-app/pull/848)
->->->[[Play] - Short Answer Response Scoring #849](https://github.com/rightoneducation/righton-app/pull/849)
->->->->[[Play/Host] - Game Flow #850](https://github.com/rightoneducation/righton-app/pull/850)
->->->->->[[Host] - Featured Mistakes #854](https://github.com/rightoneducation/righton-app/pull/854)
->->->->->->[[Play/Host] - Short Answer Formatting/Linting #855](https://github.com/rightoneducation/righton-app/pull/855)
->->->->->->->[[Play/Host] - Short Answer Response - Bug Fixes #856](https://github.com/rightoneducation/righton-app/pull/856)
->->->->->->->->[[Play/Host] - Short Answer - Helper Function Comments #857](https://github.com/rightoneducation/righton-app/pull/857)
->->->->->->->->->[[Host/Play/Networking] - Short Answer Response - Adjusts Answer Handling #858](https://github.com/rightoneducation/righton-app/pull/858)_

PR #858, opened 2023-11-06
[Host/Play/Networking] - Short Answer Response - Adjusts Answer Handling

drewjhart:
**Issue:**
During a meeting, Sinclair and I generated the following document to track how we want the Short Answer Response feature to respond:  [RightOn PWA1.1 - Short Answer Responses - Answer Handling Coordination](https://docs.google.com/document/d/1Zu405H2wDkLOMgLUQQTFTikFDPl6wL_HE5y6_alxOBw/edit)

Coming out of that, we required the following changes:

1. We needed to be able to catch if students are entering multiple numeric answers to try and game the system. For example, if the correct answer is '1200', '1200, 1100, 1500' should not also be considered correct. 

2. We need to be able to ensure that strings would be caught regardless of other words in the string. For example, if the correct answer is 'Amygdala', 'the amygdala' should also be considered correct (we're already catching lowercase). 

3. We need to be able to ensure that spaces don't impact expressions (we are already doing that) and we should consider whether or not to validate if the order of terms matters for functions. For example, if the correct answer is '4x^2+5x+4', can we catch that '5x+4x^2+4' is correct? 

4. There were some minor comments about decimals and the like. 

Code changes:

1. We flipped the order of the `normAnswers` object so that they are now sorted by answer type: 

```
export interface INormAnswer {
    [AnswerType.NUMBER]: number[];
    [AnswerType.STRING]: string[];
    [AnswerType.EXPRESSION]: string[];
}
```

This allows us to quickly detect if multiple numbers are present in an answer, and conver that to one long string instead. 

2. We used a string comparison with regex to ensure we're grabbing subwords

3. We broke out the expression detection function and are using `mathjs'` `symbolicEquality` function instead of `evaluate` to check for equivalence in functions.

4. We also cleaned up the functions in general. 

_Note: this PR is a part of a multi-part PR for the Short Answer Responses feature. The full PR structure is below:
[[Play/Networking] - Short Answer Response Base #844](https://github.com/rightoneducation/righton-app/pull/844)
->[[Host] - Short Answer Response Base #845](https://github.com/rightoneducation/righton-app/pull/845)
->->[[Networking/Host] - Short Answer Response Updates to Backend #848](https://github.com/rightoneducation/righton-app/pull/848)
->->->[[Play] - Short Answer Response Scoring #849](https://github.com/rightoneducation/righton-app/pull/849)
->->->->[[Play/Host] - Game Flow #850](https://github.com/rightoneducation/righton-app/pull/850)
->->->->->[[Host] - Featured Mistakes #854](https://github.com/rightoneducation/righton-app/pull/854)
->->->->->->[[Play/Host] - Short Answer Formatting/Linting #855](https://github.com/rightoneducation/righton-app/pull/855)
->->->->->->->[[Play/Host] - Short Answer Response - Bug Fixes #856](https://github.com/rightoneducation/righton-app/pull/856)
->->->->->->->->[[Play/Host] - Short Answer - Helper Function Comments #857](https://github.com/rightoneducation/righton-app/pull/857)
->->->->->->->->->[[Host/Play/Networking] - Short Answer Response - Adjusts Answer Handling #858](https://github.com/rightoneducation/righton-app/pull/858)_

### DIFF
--- a/host/src/lib/HelperFunctions.jsx
+++ b/host/src/lib/HelperFunctions.jsx
@@ -285,31 +285,6 @@ export const determineAnswerType = (answer) => {
 };
 
 /**
- * This provides the equality check for each normalized answer, taking the type of answer to organize the comparison
- * for more info see: https://github.com/rightoneducation/righton-app/wiki/Short-Answer-Response-%E2%80%90-Equality-Checks
- * @param {any} normValue 
- * @param {any} prevAnswerValue 
- * @param {AnswerType} prevAnswerType 
- * @returns {boolean}
- */
-export const checkEqualityWithPrevAnswer = (normAnswerType, normValue, prevAnswerValue) => {
-  switch (normAnswerType){
-    case AnswerType.STRING: // string
-    default: 
-      return normValue.includes(prevAnswerValue);
-    case AnswerType.EXPRESSION: // expression
-      try { 
-        return evaluate(normValue) === evaluate(prevAnswerValue);
-      } catch {
-        return false;
-      }
-    case AnswerType.NUMBER: // number
-      return normValue === Number(prevAnswerValue);
-  }
-};
-
-
-/**
  * this function takes a received normalized answer and loops through all normalized answers in the immediate previous answer to check for equality
  * for more info see: https://github.com/rightoneducation/righton-app/wiki/Short-Answer-Response-%E2%80%90-Equality-Checks
  * @param {string} rawAnswer
@@ -319,45 +294,13 @@ export const checkEqualityWithPrevAnswer = (normAnswerType, normValue, prevAnswe
  * @returns {boolean}
  */
 export const checkEqualityWithOtherAnswers = (rawAnswer, normAnswerType, normAnswer,  prevAnswer) => {
-   // if student enters multiple numeric answers, they will not get points
-   if (normAnswerType === AnswerType.NUMBER && normAnswer.length > 1) {
-    return false;
-  };
-
   // convert to set to optimize the equality check between two arrays
   // this prevents having to use two nested for/foreach loops etc
   const prevAnswerSet = new Set(prevAnswer);
-  console.log(normAnswer);
-  console.log(prevAnswer);
-  console.log(prevAnswerSet);
   return (
     normAnswer.some(item => prevAnswerSet.has(item)) 
     || (rawAnswer === prevAnswer.value) 
   );
-
-  // if (!isMatch) {
-  
-  // };
-  // return isMatch;
-
-  // // we have two arrays, so we need to loop through both to check for equality
-  // for (let i = 0; i < normAnswer.length; i++) {
-  //   for (let y = 0; y < prevAnswer.length; y++) {
-  //       if (checkEqualityWithPrevAnswer(normAnswerType, normAnswer[i], prevAnswer[y])) {
-  //         return true; 
-  //       }
-  //     // last ditch check on expressions in case there is a basic match that the type check in the previous conditional misses
-  //     if ((normType === AnswerType.EXPRESSION || prevAnswer.normAnswer[i].type === AnswerType.EXPRESSION) 
-  //       && (normValue.toString() === prevAnswer.normAnswer[i].value.toString())) {
-  //         return true;
-  //     }
-  //     // last ditch raw answer checks (5% === 5% enter as an expression, for instance)
-  //     if (rawAnswer === prevAnswer.value) {
-  //       return true;
-  //     }
-  //   };
-  // };
-  // return false;
 }
 
 /**
@@ -398,7 +341,7 @@ export const buildShortAnswerResponses = (prevShortAnswer, choices, newAnswer, n
   let isExistingAnswer = false;
 
   // if the answer is an expression, evaluate it and store the result
-  // this prevents do this every time we check equality with prev answers
+  // this prevents doing this every time we check equality with prev answers
   if (newAnswer.answerContent.normAnswer[AnswerType.EXPRESSION]){
     newAnswer.answerContent.normAnswer[AnswerType.EXPRESSION] = newAnswer.answerContent.normAnswer[AnswerType.EXPRESSION].map((exp) => {
       try {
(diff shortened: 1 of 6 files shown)


host/src/lib/HelperFunctions.jsx
```diff
@@ -283,31 +284,31 @@ export const determineAnswerType = (answer) => {
     return AnswerType.STRING;
   }
 };
-
 /**
```
drewjhart, line 287: annoying formatting here (probably on my end) but here's what's happening:

`checkEqualityWithPrevAnswer` is being removed. That equality check now happens in the main function `checkEqualityWithOtherAnswers` instead of offloading it to another function. 

`checkExpressionEquality` is a new function that is added that exclusively handles the checking of expressions (so that we can use `mathjs` `symbolicEqual`

so `checkEqualityWithPrevAnswer` and `checkExpressionEquality` are two separate functions with unique purposes, one isn't being replace with another

host/src/lib/HelperFunctions.jsx
```diff
@@ -318,23 +319,18 @@ export const checkEqualityWithPrevAnswer = (normValue, prevAnswerValue, prevAnsw
...
-        && (normValue.toString() === prevAnswer.normAnswer[i].value.toString())) {
-          return true;
-      }
-      // last ditch raw answer checks (5% === 5% enter as an expression, for instance)
-      if (rawAnswer === prevAnswer.value) {
-        return true;
-      }
-  };
-  return false;
+export const checkEqualityWithOtherAnswers = (rawAnswer, normAnswerType, normAnswer,  prevAnswer, correctAnswerRegex) => {
```
drewjhart, line 322: trying to optimize this as much as possible by using a set to handle the nested loops. I don't think there's a way around it for the expressions (as we need to insert that `mathjs` function to compare, so moving that off to its own function and returning from this one early if we use it

host/src/lib/HelperFunctions.jsx
```diff
@@ -318,23 +319,18 @@ export const checkEqualityWithPrevAnswer = (normValue, prevAnswerValue, prevAnsw
...
+  // this prevents having to use two nested for/foreach loops etc
+  const prevAnswerSet = new Set(prevAnswer);
+  // expression equality requires a bit more work, as we have to use mathjs to check for symbolic equality
+  if (Number(normAnswerType) === AnswerType.EXPRESSION) {
+    return checkExpressionEquality(normAnswer, prevAnswerSet);
+  }
+  return ( normAnswer.some(item => prevAnswerSet.has(item)) 
+    || (Number(normAnswerType) === AnswerType.STRING && correctAnswerRegex.test(normAnswer))
+    || (rawAnswer === prevAnswer.value) 
+  );
```
drewjhart, line 333: last two conditions here are final checks if we don't find precise equality using the some/has. This catches substrings within strings as well as a precise match from the rawAnswer

host/src/lib/HelperFunctions.jsx
```diff
@@ -347,39 +343,45 @@ export const checkEqualityWithOtherAnswers = (rawAnswer, normValue, normType, pr
  * @returns {IResponse[]}
  */
 export const buildShortAnswerResponses = (prevShortAnswer, choices, newAnswer, newAnswerTeamName, teamId) => {
+  let correctAnswer = choices.find(choice => choice.isAnswer).text;
+  const correctAnswerRegEx = new RegExp(`\\b${correctAnswer.toLowerCase()}\\b`, 'g');
```
drewjhart, line 347: we use this Regex to check substrings with the correct answer only.  So if we had the following:

correct: "Paris" 
answer 1: "The answer is Paris"
answer 2: "The answer is" 
answer 3: "the"

answer 1 will match the correct answer while answer 2 and 3 will not match the other answers

host/src/lib/HelperFunctions.jsx
```diff
@@ -347,39 +343,45 @@ export const checkEqualityWithOtherAnswers = (rawAnswer, normValue, normType, pr
...
-        isExistingAnswer = true;
-        prevShortAnswer[y].count += 1;
-        prevShortAnswer[y].teams.push({name: newAnswerTeamName, id: teamId, confidence: newAnswer.confidenceLevel});
-        break outerloop;
-      }
-    };
-  };
-      
+
+  // for each answer type in the newly submitted answer
```
drewjhart, line 370: for each type of answer, will call the equality function using the corresponding type of the previous answer.

networking/src/Models/ITeamAnswerContent.ts
```diff
@@ -12,14 +12,15 @@ export interface IExtractedAnswer {
   type: AnswerType;
 }
 export interface INormAnswer {
-  value: string | number;
-  type: AnswerType;
```
drewjhart, line 16: new `INormAnswer` per the Pr description

host/src/lib/HelperFunctions.jsx
```diff
@@ -4,7 +4,8 @@ import {
   ConfidenceLevel,
   AnswerType,
 } from '@righton/networking';
```
drewjhart, line 6: using a different `mathjs` function for expressions per PR description

networking/src/Models/ITeamAnswerContent.ts
maniramezan: hmm, can we have a mix between answer types, e.i. both number and expression types have values?

maniramezan: Also, one other way to look at this is having a based class and then subclasses for each type of answers. Then, base class can define equal method which each child class needs to override it. Let me know if you need any help or prefer an example and happy to write up something in playground. We should also restrict the types, like array of numbers as answer should be a separate answer type vs. just one answer. For example, what kind of questions have an array of expressions as answer? I can see it for string and numbers, but not expressions. It also should be a separate type which then gives us better knowledge in UI how to handle it. Like we can update the text editor in answers that after each space, shows the number as the whole with background.

maniramezan: Actually to help better understand the proposal above, here's a code snippet for it:

```typescript
enum AnswerType {
    Number,
    Numbers,
    String,
    Strings,
    Expression,
}

abstract class BaseAnswer<T> {
    answerType: AnswerType
    value: T | null = null

    constructor(answerType: AnswerType, value: T | null = null) {
        this.answerType = answerType
        this.value = value
    }

    abstract isEqualTo(answer: BaseAnswer<T>): Boolean
}

class NumberAnswer extends BaseAnswer<number> {
    isEqualTo(answer: NumberAnswer): Boolean {
        return this.value == answer.value
    }
}

let a = new NumberAnswer(AnswerType.Number, 2)
let b = new NumberAnswer(AnswerType.Number, 2)
```

PR #902, opened 2023-12-05
[Host/Networking] - Surfacing Student Thinking - Bugs

drewjhart:
This PR addresses the following bugs:

1. `host` - clears `hints` and `gptHints` at the end of a question so that they don't impact any of the following questions. 
2. `host` - rebuilds `hints` in case a teacher refreshes the page on `CHOOSE_TRICKIEST_ANSWER`


maniramezan (commented):
I think I've commented this somewhere else, but I see many styling issues, might worth to run a styling guide in `host`. I also noticed we've created the `v2` version so we can also just make sure styling is enabled and set up there and ignore the ones here

PR #907, opened 2023-12-14
Alex  play confidence card two row mobile

Alexandra-Kushnirsky:
**Issue:**
When on mobile, the radio buttons on the confidence card drop to two rows. We're looking to reduce the spacing between the buttons or the size of text when on mobile so that they stay on one line.

**Resolution:**

Opted to reduce the spacing between buttons when viewing on mobile devices. NOTE: two lines of buttons still occurs when viewing on either the Samsung S8+ or the Galaxy Fold. I can continue to look into resolving the issue on smaller devices but this may involve an additional media query so wanted to get some feedback first on if this makes sense for this small of a use case before proceeding to add additional screen size logic

Relevant code:
- In `play` > `src` > `components` > `ConfidenceMeterCard.tsx` > `line 94`- added ternary operator to check if the view is on small device and if so, decrease padding between button containers

**Preview:**
Some different mobile devices pictured below:
Screenshot 2023-12-14 at 5 19 22 PM

Screenshot 2023-12-14 at 5 19 48 PM
Screenshot 2023-12-14 at 5 20 18 PM
Screenshot 2023-12-14 at 5 20 44 PM
Screenshot 2023-12-14 at 5 21 15 PM

@drewjhart


drewjhart:
Good work on this @Alexandra-Kushnirsky. Agree that we should verify what to do on small devices. @amarax, do you have any thoughts on the below? 

To summarize, Sinclair asked us to adjust the radio buttons on the Confidence Meter so that they stay on one line. Alex has been adjusting the padding for mobile devices to achieve this but we still get the two line condition on smaller devices. I think it drops to two lines with the current code at >359px device widths. Do you know how we should handle this? We can look at decreasing the font size under the radio buttons for instance. Here's a preview with the Galaxy S8 device (360px width): 

Screenshot 2023-12-15 at 12 08 35

amarax:
@Alexandra-Kushnirsky @drewjhart I'd actually recommend to keep the spacing between the options if possible, but tighten up the padding on the outside of the card (which I would recommend to do anyway to maximise the screen). Here's what it could look like on an extreme case of 320px width for an iPhone SE 2016, which may be relevant for us as we're targeting students from lower-income households:

image
[Figma reference](https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?type=design&node-id=17548-128169&mode=design&t=NDKgdhsBpBcxLntv-4)

If this approach works well, I can provide clearer border specs, specifically for the smallest breakpoint.

Alexandra-Kushnirsky:
@amarax Got it! I made a couple of changes so that spacing in between is prioritized: mainly I decreased the padding on the edges of the card, changed the alignment from 'center' to 'space-around' and decreased the text size slightly if the user is using a smallDevice. Here are some screenshots of the result on different devices below. The Samsung S8+ looks better but galaxy fold still looks a bit awkward so I can continue to tinker around if you feel it is necessary:
Screenshot 2023-12-20 at 5 12 47 PM
Screenshot 2023-12-20 at 5 10 57 PM
Screenshot 2023-12-20 at 5 11 25 PM
Screenshot 2023-12-20 at 5 11 54 PM
Screenshot 2023-12-20 at 5 12 06 PM
Screenshot 2023-12-20 at 5 12 29 PM

drewjhart:
@amarax mentioned that he's reviewing some broader UI things before commenting on these issues but that we shouldn't let him hold us up. I'm going to merge this and then he can follow up with comments next week. @Alexandra-Kushnirsky

Issue #906, opened 2023-12-14
[Host] - Scroll screen on graph click


drewjhart:
# Issue: General #

## Overview ##
Currently, we are scrolling to the top of the screen when a component is clicked. See the video below:

https://github.com/rightoneducation/righton-app/assets/79417944/dab39649-5590-4418-a749-395b0024d46f

We are looking to clean up this behaviour so it doesn't cause such dramatic changes across screens. Instead of scrolling to the top of the screen, we are looking for it to scroll to the top of the "Showing Players Who Answered"  component:
![image](https://github.com/rightoneducation/righton-app/assets/79417944/b4a67631-941d-492e-b1e2-703603030506)

## Relevant Code: ##
`host`
1. `GameInProgress` - `handleGraphClick` Line 111. Uses `refs` that are assigned to each component to govern scroll behaviour when graph is clicked.

PR #908, opened 2023-12-15
[Host] - Scroll screen on graph click

Alexandra-Kushnirsky:
**Issue:**
Currently, we are scrolling to the top of the screen when a component is clicked: https://github.com/rightoneducation/righton-app/issues/906 

**Resolution:**
Changed ref to point to subcomponent containing text that says "Showing players who answered:"

Relevant code:
- In `host` > `src` > `components` > `GameInProgressContentSwitch.jsx` > lines `57-82`: removed ref assignment from parent div and added new div wrapper with ref assignment to component we want to scroll to on `lines 70, 71`: 
`<div id="responses-scrollbox"
              ref={responsesRef}>
              <SelectedAnswer...`

**Preview:**
In the below video, we can see the scroll behavior on bar click has changed so that we focus on the subcomponent containing text that says "Showing players who answered:"

https://github.com/rightoneducation/righton-app/assets/106450593/7f41fb3a-89a6-4241-94dd-2f00b4516ec0

@drewjhart

### DIFF
--- a/host/src/components/GameInProgressContentSwitch.jsx
+++ b/host/src/components/GameInProgressContentSwitch.jsx
@@ -14,49 +14,47 @@ import FeaturedMistakes from "./FeaturedMistakes";
 import PlayerThinking from "./PlayerThinking/PlayerThinking";
 import PlayerThinkingSelectedAnswer from './PlayerThinking/PlayerThinkingSelectedAnswer';
 
-export default function GameInProgressContentSwitch ({ 
-    questions, 
-    questionChoices,
-    data,
-    graphClickInfo,
-    responsesRef,
-    handleGraphClick,
-    currentQuestionIndex, 
-    answers, 
-    totalAnswers, 
-    numPlayers, 
-    statePosition, 
-    questionCardRef, 
-    gameAnswersRef,
-    confidenceCardRef,
-    featuredMistakesRef,
-    correctChoiceIndex,
-    currentState,
-    isConfidenceEnabled,
-    handleConfidenceSwitchChange,
-    teamsArray,
-    confidenceData,
-    isShortAnswerEnabled,
-    handleShortAnswerChange,
-    shortAnswerResponses,
-    handleOnSelectMistake,
-    hintCardRef,
-    isHintEnabled,
-    handleHintChange,
-    hints,
-    gptHints,
-    hintsError,
-    isHintLoading,
-    handleProcessHints
-  }) {
+export default function GameInProgressContentSwitch({
+  questions,
+  questionChoices,
+  data,
+  graphClickInfo,
+  responsesRef,
+  handleGraphClick,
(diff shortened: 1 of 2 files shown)


drewjhart:
Looks like it's working as it should, nice. Would you be able to do a video where there are like 10 players (it will be super annoying to generate)? I just want to test the scroll when that list of players doesn't fit on the screen

Alexandra-Kushnirsky:
https://github.com/rightoneducation/righton-app/assets/106450593/94ae50ab-3980-4046-897b-1e575faeb789

drewjhart:
Thanks for updating this video, Alex - even if it was a pain. Is there a way to get the 'Players Who Have Answered' component to the top of the screen when a user clicks the bar graph? I think right now it's just ensuring that the component is entirely on the screen but it would be good to get it at a consistent place.

Alexandra-Kushnirsky:
No problem at all! I changed `responsesRef.current.scrollIntoView({ behavior: 'smooth'});` to `responsesRef.current.scrollIntoView({ behavior: 'smooth', block: 'start' });` on `line 154` of `host`>`src`>`pages`>`GameInProgress.jsx` so that the component we're targeting scrolls to the highest permittable vertical position on the screen. Basically, this means if there are only a few players that selected that answer, it won't go all the way to the top since there would be blank space underneath but if we want this to be the case, I can find a different solution. Here is a video with the new behavior below. You can see both cases where (1) there are many players who selected an answer and thus the dropdown takes up a significant portion of the screen and (2) the opposite case where only a few players selected an answer:


https://github.com/rightoneducation/righton-app/assets/106450593/d61cfec1-dfa0-446e-a50f-2ba2cb1f0841

drewjhart:
> No problem at all! I changed `responsesRef.current.scrollIntoView({ behavior: 'smooth'});` to `responsesRef.current.scrollIntoView({ behavior: 'smooth', block: 'start' });` on `line 154` of `host`>`src`>`pages`>`GameInProgress.jsx` so that the component we're targeting scrolls to the highest permittable vertical position on the screen. Basically, this means if there are only a few players that selected that answer, it won't go all the way to the top since there would be blank space underneath but if we want this to be the case, I can find a different solution. Here is a video with the new behavior below. You can see both cases where (1) there are many players who selected an answer and thus the dropdown takes up a significant portion of the screen and (2) the opposite case where only a few players selected an answer:
> 
>  Screen.Recording.2023-12-20.at.4.01.36.PM.mov

Sick, looks good, nice one. @amarax and @sincwu, is this the kind of behaviour you're looking for in the scroll on the click of a bar?

drewjhart:
@amarax mentioned that he's reviewing some broader UI things before commenting on these issues but that we shouldn't let him hold us up. I'm going to merge this and then he can follow up with comments next week. @Alexandra-Kushnirsky @sincwu

PR #909, opened 2023-12-18
[Host] - Host_V2 Base Framework

drewjhart:
The design team has previously provided an updated design for `host`. This PR acts as a base for the framework for this new design and copies all packages and components from `play`.

For information on project structure and the dependencies, see the readme.


drewjhart:
@maniramezan, Alex and Josephina have joined us for about a month and a half over the holidays. Sinclair and I thought that it'd be good to have them work on implementing the updated design for `host`. This is a base framework to get them started. I don't think you need to actually review this PR (it's the same set up as `play`) but wanted to keep you in the loop.

host_v2/.gitignore
maniramezan, line 1: Missing Windows hidden file, `thumbs.db`

PR #921, opened 2023-12-19
[Central] - Pagination from new Api

drewjhart:
WIP

PR #919, opened 2023-12-19
[Central] - Multitable - Questions

drewjhart:
WIP

PR #922, opened 2023-12-20
[Central] - ApiClient updates

drewjhart:
Hi @maniramezan, I'm still working through these PRs as I update the backend and connect `central`. That being said, I wanted to get your eyes on these earlier, instead of waiting until I'm finished with everything and blasting you with PRs. 

I don't think you need to review all of these PRs just yet but this individual PR will give you a good indication of what I'm doing. The main idea here is to take the backend configuration that I developed and work it into the structure that you developed in your Migrate to DynamoDB PR from the summer. I think this should give you a good indication of where this whole thing is heading. 

Done:

- Adjust graphql.schema to include `central` level data objects. These are `GameTemplate`, `QuestionTemplate` and `GameQuestions` (the connection between the two templates)
- Add Lambda function via `amplify add function` to create a `GameSession` with `Questions` from a `GameTemplate` and any connected `QuestionTemplates`
- Adjust queries to properly paginate the results from either `GameTemplate` or `QuestionTemplate` via `nextToken` and `limit`
- Adjust `central` to include both a `Explore Game` and `Explore Question` option as well as a `Game Maker` and `Question Maker` option. Display the number of questions in each game and the number of games each question is deployed in via the `GameQuestions` connection. UI for `central` generally is under review. This is just to allow Sinclair and Yong Lin (and anyone else) to demo our functionality as they continue desigining.
- Split apart the API per your PR from the summer for clarity (it really helps!)

Still to Do:
- Finish up writing the `QuestionTemplate` functions on the client side of `central`
- Finish up writing the `GameQuestion` functions on the client side of `central`
- Adjust `host` and `play` to reflect new API structure
- Auth layer of API
- Renaming `GameSession` to just `Game` (so it follows Questions, etc)
- Rename `GameQuestions` to `GameTQuestionsT` to underscore they are templates
- General code clean up

## Summary ## 

Updates API Client to follow structure set out in Mani's:
[Migrate to DynamoDB #778](https://github.com/rightoneducation/righton-app/pull/778) 

while using the amplify backend and central updates established in:
[[Networking] - Updates GameTemplate and QuestionTemplates into DynamoDB #903](https://github.com/rightoneducation/righton-app/pull/903)
-> [[Central] - Multitable - Questions #919](https://github.com/rightoneducation/righton-app/pull/919)
->-> [[Central] - CreateGameSession Lambda Function #920](https://github.com/rightoneducation/righton-app/pull/920)
->->-> [[Central] - Pagination from new Api #921](https://github.com/rightoneducation/righton-app/pull/921)


networking/src/Models/AWS/AWSModelGameQuestionConnection.ts
```diff
@@ -0,0 +1,19 @@
+import { AWSQuestionTemplate } from "./AWSQuestionTemplate";
+import { AWSGameTemplate } from "./AWSGameTemplate";
+
+export type AWSModelGameQuestionConnection = {
+  items: Array<AWSGameQuestions>;
+  __typename: string;
```
maniramezan, line 6: Same as other one, I don't think we need `__typename` here, and in any other models on client side

networking/src/Models/AWS/AWSQuestionTemplate.ts
```diff
@@ -0,0 +1,19 @@
...
+    version?: number | null,
+    choices?: string | null,
+    instructions?: string | null,
+    answerSettings?: string | null,
+    domain?: string | null | undefined,
+    cluster?: string | null | undefined,
+    grade?: string | null | undefined,
+    standard?: string | null | undefined,
+    imageUrl?: string | null | undefined,
+    gameTemplates?:  AWSModelGameQuestionConnection | null,
```
maniramezan, line 16: nit: extra space after `:`

networking/src/Models/AWS/AWSTeamAnswer.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type AWSTeamAnswer = {
+  id: string
+  questionId?: string | null
+  isChosen: boolean
+  isTrickAnswer: boolean
+  text?: string | null
+  awsAnswerContents?: string | null
+  createdAt?: string
+  updatedAt?: string
```
maniramezan, line 11: These shouldn't be nullable for this model and other `AWSXXX` models

central/src/lib/API/gamequestions.ts
maniramezan, line 1: File name to be pascal based and should also include `APIClient` in the name as well. Any reason for using lower-case naming for all the files here and not include the `APIClient` in the name?

maniramezan, line 1: For all the methods here, probably worth to have them throw instead of returning `null`. That way, the caller can use the same pattern for promises to capture the error and show an appropriate error message in UI. The issue with the current code is that the caller needs to always check for `null` and can't use promises to `catch` to capture it.

Also, no need to store it locally in a variable before returning it, can just return it to reduce the boilerplate code. I know writing these can be too much of repeating same pattern, if that becomes a pain, we can look into codegen alternative to autogenerate all of these

For rest of api ones here, same comment.

networking/src/Parsers/TeamAnswerParser.ts
```diff
@@ -0,0 +1,86 @@
+import { isNullOrUndefined } from "../IApiClient";
```
maniramezan, line 1: With removing `APIClient`, there'll be no usage for `IApiClient` as well, might worth to move this method to a global file. I believe we can create `global.ts` to be publicly available.

networking/src/APIClients/interfaces/IGameQuestionsAPIClient.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface IGameQuestionsAPIClient {
+  createGameQuestions(
+    createGameQuestionsInput: CreateGameQuestionsInput
+  ): Promise<IGameQuestions | null>;
+
+  getGameQuestions(
+    id: string
+  ): Promise<IGameQuestions | null>;
+
+  deleteGameQuestions(
```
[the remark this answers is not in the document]
drewjhart: fixed

networking/src/APIClients/interfaces/IGameQuestionsAPIClient.ts
```diff
@@ -0,0 +1,21 @@
+import { IGameQuestions } from "../../Models";
+import { CreateGameQuestionsInput } from "../../AWSMobileApi";
+
+export interface IGameQuestionsAPIClient {
+  createGameQuestions(
```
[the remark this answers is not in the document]
drewjhart: fixed

networking/src/Models/AWS/AWSModelGameQuestionConnection.ts
drewjhart: fixed

networking/src/Models/AWS/AWSQuestionTemplate.ts
drewjhart: good eyes 👀 - fixed

central/src/lib/API/gamequestions.ts
drewjhart: fixed

networking/src/Models/AWS/AWSTeamAnswer.ts
drewjhart: fixed

networking/src/APIClients/interfaces/IBaseAPIClient.ts
```diff
@@ -0,0 +1,5 @@
+export enum Environment {
+  Staging = "staging",
+  Developing = "developing",
+  Testing = "testing"
```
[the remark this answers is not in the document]
drewjhart: This is indeed janky. "Staging" is coming from the preexisting setup with the API Gateway for CreateGameSession lambda. So 'Staging' here is actually our production env. I'd like to talk with you a bit further about naming here though because one thing is hanging me up: 

I am assuming the structure for the above environments is: `developing` -> `testing/staging` -> `production`

Where `developing` is often changing through development, `testing/staging` is slightly more stable, and then `production` only changes on releases.

However, for our client side code, I think this owuld mean that our `dev` branch uses the `testing/staging` environment (so that the branch doesn't break when the backend `env` changes). It feels kind of weird to have this mismatch, where `dev` doesn't use the `developing` environment but is that as intended? 

If we could spend 5 minutes talking through this on a meeting, that would be amazing

central/src/lib/API/gamequestions.ts
drewjhart: I thought I was following some preexisting code, but you're right - it doesn't make much sense. renamed folders and files to follow `networking`

PR #944, opened 2024-01-08
[Backend]- Consolidate Team and Correct Answer Classes

drewjhart:
Resolves comments from [this thread](https://github.com/rightoneducation/righton-app/pull/875#discussion_r1445108979) in [[Play/Networking] - Short Answer Response Base #875](https://github.com/rightoneducation/righton-app/pull/875)


networking/src/Models/AnswerClasses.ts
```diff
@@ -232,10 +153,20 @@ export class NumberAnswer extends TeamAnswerClass<number> {
     }
     return false;
   }
+
```
drewjhart, line 156: Migrating answer validation to `AnswerClasses`

Issue #914, opened 2023-12-19
[Host_v2] - Responsive Layout for Cards


drewjhart:
# Issue: General #

## Overview ##
The design notes that there are three possible configurations for host, based on screen width:
Small:
<img width="283" alt="Screenshot 2023-12-19 at 11 33 50" src="https://github.com/rightoneducation/righton-app/assets/79417944/680a2da6-9071-4a9f-a739-93940de35221">
Medium:
<img width="409" alt="Screenshot 2023-12-19 at 11 34 12" src="https://github.com/rightoneducation/righton-app/assets/79417944/08e2aeb3-b88e-479a-bb45-5a966ceafc28">
Large:
<img width="328" alt="Screenshot 2023-12-19 at 11 34 24" src="https://github.com/rightoneducation/righton-app/assets/79417944/5a55b72a-2746-4b7e-a765-960f213b974b">

Notes: 
- Note that for small and medium devices, the other columns are indicated partly off screen so that users know that they can swipe over to see them.
- This is similar to what we are doing on `play` with the [Swiper](https://swiperjs.com/) for the carousels, but with an additional configuration so that we can swipe across three columns. 
- Please follow the styling for the pagination bullets in `play` rather than the explicit design. I asked Yong Lin for a compromise on these because there is a limited amount of styling you can easily do via Swiper on the bullets.

## Figma Links ##

- [Small Devices (older design so don't worry about subtle changes)](https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?type=design&node-id=14571-126301&mode=dev)
- [Medium Devices (most recent design)](https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?type=design&node-id=15088-135077&mode=dev) 
- [Large Devices (most recent design)](https://www.figma.com/file/PadzU8GqyhFLCjIt7rQkpE/RightOn-Apps?type=design&node-id=15090-142357&mode=dev)

## Relevant Code ##
- Swiper JS setup - play/src/pages/gameinprogress/ChooseAnswer.tsx (includes the flag that enables the swiper or not)

PR #945, opened 2024-01-08
Alex  host v2 responsive layout

Alexandra-Kushnirsky:
**Issue:**
Addresses items for git issue #914: create responsive layout for host_v2 cards

**Resolution:**

- Created three separate components to be rendered conditional on screen size using a media query (smallDevice, mediumDevice, and largeDevice)
- Adapted Swiper, ScrollBoxStyled from play to enable swiping across carousel columns for small and medium devices
- This can be found in `host_v2>src>components>PlaceHolderContentArea.tsx`

Relevant code:
- `host_v2>src>components>PlaceHolderContentArea.tsx` - Bulk of the code pertaining to this PR is in this file. It contains the media queries, the Swiper carousel components for each respective screen, and lastly a conditional rendering of the appropriate UI component based on the media query
- `host_v2>src>lib>styledComponents>layout>ScrollBoxStyled.tsx` - added this file (which is essentially the same in play outside of the height calculation since the themes vary between host_v2 and play) to allow for vertical scroll on Swiper slides without moving the page horizontally
- `host_v2>src>lib>styledComponents>layout>BodyContentAreasStyled.tsx` - added new `const BodyContentAreaTripleColumnStyled` on `lines 4-18` with a larger max width to account for styling on large devices that calls for 3 columns without a swiper
- host_v2>src>lib>styledComponents>PaginationContainerStyled.tsx` - Added by Drew to enable pagination for Swiper

**Preview:**

Here is how the responsive layout UI looks on different devices: 

https://github.com/rightoneducation/righton-app/assets/106450593/44e980d4-f17f-45a6-b1d3-60c21e50075b

@amarax , @drewjhart

### DIFF
--- a/host_v2/package.json
+++ b/host_v2/package.json
@@ -23,8 +23,10 @@
     "react-i18next": "^13.5.0",
     "react-router-dom": "^6.21.0",
     "react-scripts": "5.0.1",
+    "react-spring": "^9.7.3",
     "storybook": "^7.6.5",
     "styled-components": "^6.1.1",
+    "swiper": "^11.0.5",
     "typescript": "^4.9.5",
     "uuid": "^9.0.1",
     "web-vitals": "^2.1.4"
--- a/host_v2/src/components/PlaceholderContentArea.tsx
+++ b/host_v2/src/components/PlaceholderContentArea.tsx
@@ -34,17 +34,19 @@ export default function PlaceholderContentArea({ }: PlaceholderContentAreaProps)
   const mediaQueryLarge = window.matchMedia(`(min-width: ${theme.breakpoints.values.lg}px)`);
   const isLargeScreen = mediaQueryLarge.matches;
 
-  // TODO: 
-  //      -update scroll behavior if necessary (after getting feedback from design)
   const largeScreen =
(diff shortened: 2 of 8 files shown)


drewjhart:
Nice job on this, @Alexandra-Kushnirsky  - this is looking really good. One thing I noticed was that on the small screen size I think a fixed width is set. This means that when we hit that breakpoint, the swiper column drops to that size even if the screen size is great. This results in a gap between the partially shown slide and the edge of the screen. Here are a couple videos to illustrate this:

`play` (intended behaviour):

https://github.com/rightoneducation/righton-app/assets/79417944/0ebde355-e036-4fce-8711-4fb174faf822

Note that the partially shown card stays at the edge of the scren.

`host_v2` (current branch)

https://github.com/rightoneducation/righton-app/assets/79417944/3b68ce11-a6f3-4816-b149-026ea7628f35

Note the gap that is produce between the partially shown card and the edge of the screen when screen size is small (looks like it might be working ok on medium size)

Otherwise, looking good though!

drewjhart (approved):
Nice!

PR #1082, opened 2024-05-09
[Backend] - Public/Private Ownership

drewjhart:
## Background: ##
After discussing with the team (captured in the wiki [here](https://github.com/rightoneducation/righton-app/wiki/Backend-%E2%80%90-Authentication-and-Authorization#publicprivate)), we have decided to implement separate tables for Public and Private `GameTemplates` and `QuestionTemplates`. This PR and the following one implement that strategy.

They break down per:
[[Backend] - Public/Private Ownership #1082](https://github.com/rightoneducation/righton-app/pull/1082)
-> [[Backend] - Public/Private Ownership - Central UI Changes #1096](https://github.com/rightoneducation/righton-app/pull/1096)
->-> [[Backend] - Public/Private - Bug Fixes #1101](https://github.com/rightoneducation/righton-app/pull/1101)

 This PR (#1082) is focused around `networking` changes that provide the required updates to the `schema.graphql` and the associated `APIClients`. The second PR (#1096) is focused around UI updates to ensure that `central` is functional. The last PR (#1101) resolves some bugs.

## Updates: ##

To summarize the work here, it falls into three main buckets:

1. `schema.graphl` Updates:
The `schema.graphql` file has been updated to include `PublicGameTemplate`, `PrivateGameTemplate`, `PublicQuestionTemplate`, and `PrivateQuestionTemplates`. Authorization has been set as follows:

`PublicGameTemplate`/`PublicQuestionTemplate`:
```graphql
@auth(
  rules: [
    { allow: public, operations: [read], provider: iam  },
    { allow: owner, provider: userPools },
    { allow: private, operations: [read], provider: userPools },  
  ]
)
``` 

`PrivateGameTemplate`/`PrivateQuestionTemplate`:
```graphql
@auth(
  rules: [
    { allow: owner, provider: userPools }
  ]
)
```

As such, Public templates are able to be read by everyone but are only able to be edited by their owners while Private templates are only able to be read or written by their owners. 

One other note, each template has been updated with a new field:
```graphql
owner: String
```

It's important that this field is both: 1) added and 2) remains optional. `Amplify` will write to this field upon initial creation to infill with user information. However, prior to creation, we will send in a null value. Therefore, we need to ensure that the field is present, and also able to be null. Some hoops `Amplify` makes us jump through.

2. `I<Type>TemplateAPIClient` updates:
`Amplify` autogenerates our queries and resolvers in a particular way. It can generate something like this: `publicGameTemplateQuery` or `getPublicGameTemplateQuery` or `PublicGameTemplateQuery`, so there are a series of names with different casing and types associated with the autogen. I've rewritten our interfaces as a series of objects and a map so that we can process down the autogenerated code into something that we can easily handle in our `APIClients` without too much clutter. Here's a sample, taking `PublicGameTemplate` as a case study:

```ts
export interface IPublicGameTemplate {
  create: {
    input: CreatePublicGameTemplateInput;
    variables: CreatePublicGameTemplateMutationVariables;
    query: CreatePublicGameTemplateMutation;
  },
  get: {
    variables: GetPublicGameTemplateQueryVariables;
    query: GetPublicGameTemplateQuery;
  },
  update: {
    input: UpdatePublicGameTemplateInput;
    variables: UpdatePublicGameTemplateMutationVariables;
    query: UpdatePublicGameTemplateMutation;
  },
  delete: {
    input: DeletePublicGameTemplateInput;
    variables: DeletePublicGameTemplateMutationVariables;
    query: DeletePublicGameTemplateMutation;
  },
  list: {
    variables: ListPublicGameTemplatesQueryVariables;
    query: ListPublicGameTemplatesQuery;
  }
}
```

Interface strongly types each of the input, variable, and query autogen'd code that we import. 

```ts
export const gameTemplateRuntimeMap = {
  Public: {
    create: {
      queryFunction: createPublicGameTemplate,
    },
    get: {
      queryFunction: getPublicGameTemplate,
    },
    update: {
      queryFunction: updatePublicGameTemplate,
    },
    delete: {
      queryFunction: deletePublicGameTemplate,
    },
    list: {
      queryFunction: {
        default: listPublicGameTemplates,
        byDate: publicGameTemplatesByDate,
        byGrade: publicGameTemplatesByGrade,
        byQuestionTemplatesCount: publicGameTemplatesByPublicQuestionTemplatesCount
      },
    },
  },
  Private: {
    ... there is a private version too
    },
  }
}
```
Runtime map provides an object we can reference for the autogen'd queries (that aren't types). 

We can then make use of this like so:

```ts
export type GameTemplateType<T extends PublicPrivateType> = T extends 'Public' ? IPublicGameTemplate : IPrivateGameTemplate;

export interface IGameTemplateAPIClient {
  createGameTemplate<T extends PublicPrivateType>(
    type: T,
    createGameTemplateInput: GameTemplateType<T>['create']['input'] | IGameTemplate
  ): Promise<IGameTemplate>;
 ```

We can use `<T>` to denote the `public`/`private` and we can do `['create']['input']` to grab the right typing for associated query input.

3. `<Type>TemplateAPIClient` Updates:
The actual APIClients then need to be updated to conform to the new interfaces. This means editing all the functions to accurately provide `Public/Private` switches and to ensure strong typing. This looks like the following:

```ts
 async createGameTemplate<T extends PublicPrivateType>( 
    type: T,
    createGameTemplateInput: GameTemplateType<T>['create']['input'] | IGameTemplate
  ): Promise<IGameTemplate> {
    const variables: GraphQLOptions = { input: createGameTemplateInput as GameTemplateType<T>['create']['input']};
    const queryFunction = gameTemplateRuntimeMap[type].create.queryFunction;
    const createType = `create${type}GameTemplate`;
    const gameTemplate = await this.callGraphQL<GameTemplateType<T>['create']['query']>(
        queryFunction,
        variables
    ) as { data: any};
    if (
        isNullOrUndefined(gameTemplate?.data)
    ) {
        throw new Error(`Failed to create game template.`)
    }
    return GameTemplateParser.gameTemplateFromAWSGameTemplate(gameTemplate.data[createType] as AWSGameTemplate, type)
  } 
  ```

### CODE REVIEW

File: central/src/lib/API/GameQuestions.ts [hunk 1]
```diff
@@ -1,22 +1,37 @@
...
   IGameQuestion,
-  CreateGameQuestionsInput
+  CreatePublicGameQuestionsInput,
+  CreatePrivateGameQuestionsInput,
+  PublicPrivateType
 } from '@righton/networking';
 
-export const createGameQuestions = async (apiClients: IAPIClients, createGameQuestionsInput: CreateGameQuestionsInput): Promise<IGameQuestion | null> => {
+export const createGameQuestions = async (
+  type: PublicPrivateType,
```
drewjhart, line 10: all calls to the backend now include a `PublicPrivateType` that describes if the call is for the Public or Private DynamoDB table.

central/src/lib/HelperFunctions.tsx
```diff
@@ -1,9 +1,13 @@
-import { GameTemplate, QuestionTemplate } from '@righton/networking';
```
drewjhart, line 1: update helper functions to include type

networking/amplify/backend/api/mobile/schema.graphql
```diff
@@ -1,6 +1,6 @@
 type Mutation {
   createGameSessionFromTemplate(input: CreateGameSessionFromTemplateInput!): String 
-  @function(name: "createGame-${env}") 
+  @function(name: "createGameTest-${env}")
```
drewjhart, line 3: Amplify issue that I will address when I merge all of these branches in (final name will be `createGame`

networking/amplify/backend/api/mobile/schema.graphql
```diff
@@ -9,28 +9,20 @@ type Mutation {
...
 
-input CreateGameSessionFromTemplateInput {
-  gameTemplateId: ID!
-}
-
-# define a GameTemplate in amplify and graphql to list all gametemplates sorted by updatedAt
-type GameTemplate @model
+type PublicGameTemplate @model
 @auth(
   rules: [
```
drewjhart, line 14: Here is where we provide edit access to `owner` exclusively

networking/amplify/backend/api/mobile/schema.graphql
```diff
@@ -9,28 +9,20 @@ type Mutation {
...
-    { allow: private, operations: [read, create, update, delete], provider: userPools }
+    { allow: owner, provider: userPools },
+    { allow: private, operations: [read], provider: userPools },  
   ]
 )
 @aws_iam
 @aws_cognito_user_pools
 {
   id: ID!
+  owner: String
```
drewjhart, line 24: important to add owner in here as an optional parameter. On creation, amplify populates this field with the owner information, but it needs to be specified as a field that can be null upon initial creation. Ugh

networking/amplify/backend/api/mobile/schema.graphql
```diff
@@ -9,28 +9,20 @@ type Mutation {
...
   title: String!
-  owner: String! @index(
```
drewjhart, line 29: indexing off owner blows up a lot of the background amplify processes, and isn't required as the @auth params specified above will take care of the query results etc

drewjhart: More clearly, we won't need to explicitly return values by owner in central, because we are using the @auth operations (read, write, etc) to do that for us.

networking/amplify/backend/api/mobile/schema.graphql
```diff
@@ -40,33 +32,68 @@ type GameTemplate @model
...
-input GameTemplateInput {
+input PublicGameTemplateInput {
   title: String
   owner: String
   version: Int
   description: String
 }
 
-type QuestionTemplate @model 
+type PrivateGameTemplate @model
```
drewjhart, line 49: private templates only have owner level access

PR #1116, opened 2024-06-26
[Host_v2] - Confidence Level Graph

drewjhart:
WIP


play/src/pages/GameInProgress.tsx
```diff
@@ -197,15 +197,17 @@ export default function GameInProgress({
...
       const correctAnswer = ModelHelper.getCorrectAnswer(currentQuestion)?.text ?? null;
       if (correctAnswer){
         if (isAnswerNumeric(answer.answer))
           answer.isCorrect = answer.answer.isEqualTo([Number(correctAnswer)]) as boolean; // eslint-disable-line 
-        else 
+        else {
           answer.isCorrect = answer.answer.isEqualTo([correctAnswer]) as boolean; // eslint-disable-line 
-        if (isAnswerMultiChoice(answer.answer))
+        }
+        if (isAnswerMultiChoice(answer.answer)){
```
maniramezan, line 207: nit: Space before `{`, looks like the linter might not correcting this on save

play/src/pages/GameInProgress.tsx
```diff
@@ -197,15 +197,17 @@ export default function GameInProgress({
...
           answer.answer.multiChoiceCharacter = multiChoiceCharacter; // eslint-disable-line
+          answer.answer.rawAnswer = multiChoiceCharacter; // eslint-disable-line
```
maniramezan, line 209: Why linter is disabled? Which rule are we breaking?

host_v2/src/components/ConfidenceGraph/ConfidenceResponseDropdown.tsx
```diff
@@ -133,43 +118,24 @@ export default function ConfidenceResponseDropdown({
...
-    incorrectPlayers.sort((a, b) => {
-      // sort incorrect by answer frequency, then alphabetically
-      const freqDifference =
-        answerFrequency[b.answer] - answerFrequency[a.answer];
-      return freqDifference !== 0
-        ? freqDifference
-        : a.name.localeCompare(b.name);
-    });
+    correctPlayers.sort((a, b) => a.team.localeCompare(b.team));
+    incorrectPlayers.sort((a, b) => a.team.localeCompare(b.team));
```
maniramezan, line 131: nit: Since these two are the same sort logic, we can create a local variable for the sort method and then pass it here as input

maniramezan (approved):

PR #1141, opened 2024-07-24
[Host_v1] - Timer Synch via host and play

drewjhart:


kellycouvrette:
This branch syncs up the host timer and the play timer using the code from the host_v2 and play syncing. 
Main changes involved-
- GameSessionContainer.tsx - added in the allotted time and calculateCurrentTime functions from host_v2. passed in these values to GameInProgress instead of the old timer implementation
- GameInProgress.tsx - handles the footerOnClick so we can store the startTime of each question in the backend
- StudentViews.tsx - does the same as above, but for when the question is started from the studentViews page (start of phase 2 questions)

PR #1149, opened 2024-07-30
[Host_v2] - Fixes

drewjhart:


host_v2/src/components/FeaturedMistakes.tsx
```diff
@@ -69,16 +67,19 @@ export default function FeaturedMistakes({
   const radioButtonText1 = 'Use the top 3 answers by popularity';
```
drewjhart, line 67: removes a local `useState` variable for mistakes and instead derives entirely off of `localHostTeamAnswers` using context and dispatch

PR #1248, opened 2024-10-15
[Central_v2] - Top Level Organization/Container Setup

drewjhart:
This PR organizes the app code generally to support the development of `central_v2`. This includes:

1. Integrates `react-router-dom` to add routes for each of the different tabs
2. Adds AppContainer as a wrapper that always includes the header
3. Adds AppSwitch that works as the switch based on the route match
4. Starts to integrate `styled-components` organization, moving a lot of styled components that are located at their components into the `lib` folder


drewjhart:
Hey @maniramezan, I'm tagging you in a couple PRs for `central_v2` for review. Kelly was working on some initial `central_v2` stuff last month but these two PRs represent the initial set up for the app as a whole. They include some reorganization around containers/switches etc, and then the integration of `Context` as well as the creation of a `CentralDataManagerAPIClient` (similiar to what we did for  host and play).

Muhammad is joining the team this week so I'm going to be merging these PRs into dev so we can all work off a consistent base. That being said, I will tag you in the PRs as they come up so you can see what we are doing and if you have any comments, I can address them in subsequent PRs.

Thanks a ton and let me know if this works for you!

PR #1403, opened 2025-01-12
Muhammad bucket

CCosmoe:
**Issue:**

- Allow customers to upload images.

**Resolution:**

The task of allowing teachers to add front and back images to their account has been implemented through the use of s3 buckets. The user Signs up for their account as usual by putting in their first name, last name, email, image,s etc. Once the user presses the next step, all the information except the teacher images gets sent to the DynamoDB table. Teacher Images stay on the client side for now. The user is then directed to the Confirmation page. This is where the user gets the code from the email that they used and inputs it on that page. Once the user presses verify their account is authenticated the teacher images get sent to the s3 buckets. This is done to prevent users from uploading any malicious data.

Relevant code:

- SignUp.tsx: This page renders the SignUp page. This is what the users sees once they want to register for an account. In this component, the user information gets sent to the DynamoDb table once they press next step. This is also where we switch between the Confirmation.tsx page and SignUp.tsx page. Code can be found in central_v2 --> src --> pages.

- Confirmation.tsx: In this component the user is able to input a code that was sent to their email to authenticate them. This is also where the teacher images are sent to the s3 buckets. Code can be found in central_v2 --> src --> pages.

- UserAPIClients.ts: This is where we made the function uploadTeacherID. The purpose of this function is to take the inputs(teacher images) and call Graph QL with the auto-generated queries from modifying schema.graphQL. This function is used on the client side in Signup.tsx component in the handlerImageUpload function. Code can be found in networking --> src --> APIClients --> user --> UserAPIClients.

- Index.mjs: This is the lamda function that is called once we upload teacher images. We take the teacher image and insert them into the s3 bucket. Code can be found in networking --> amplify --> backend --> function --> teacherIdAuth --> src. 

- schema.graphql: This is where we add new tables and attach lambda functions. In this we added teacherIdAuth a type mutation that has the teacherIdAuth function(lambda) attach to it. When this function is called it triggers Index.mjs file as discussed above.

Below are the images that show what the user sees for both pages.

image
image


drewjhart:
merged via: https://github.com/rightoneducation/righton-app/pull/1434

PR #1530, opened 2025-03-26
[Central_v2] - Substitutes useState for Context/Reducer for all central data 

drewjhart:


central_v2/src/App.tsx
```diff
@@ -9,26 +9,16 @@ import { useAPIClients, Environment, AppType } from '@righton/networking';
 import { GoogleOAuthProvider } from '@react-oauth/google';
 import { ThemeProvider, StyledEngineProvider } from '@mui/material/styles';
 import { APIClientsContext } from './lib/context/APIClientsContext';
-import UserProfileReducer from './lib/reducer/UserProfileReducer';
 import Theme from './lib/Theme';
 import AppSwitch from './switches/AppSwitch';
 import CreateQuestionLoader from './loaders/CreateQuestionLoader';
```
drewjhart, line 14: I'm removing UserProfileContext and replacing it with CentralDataContext, which will incoporate UserProfile as well as all other state variables that lean on an API request

central_v2/src/components/button/favouritebutton/FavouriteButton.tsx
```diff
@@ -19,14 +18,17 @@ export default function FavouriteButton({
 }: FavouriteButtonProps) {
   const [isLoading, setIsLoading] = useState<boolean>(false);
   const apiClients = useTSAPIClientsContext(APIClientsContext);
-  const userProfile = useUserProfileContext(UserProfileContext);
-  const userProfileDispatch = useUserProfileDispatchContext(UserProfileDispatchContext);
-  const isFavorite = userProfile?.favoriteGameTemplateIds?.includes(id) ?? false;
+  const centralData = useCentralDataState();
```
drewjhart, line 21: this is what it looks like to use the context down in the component tree. You can think of this like 

const `[state, setState] = useState();`

where `centralData` is the equivalent of `state` and `centralDataDispatch` is the equivalent of `setState`

central_v2/src/components/librarytabs/LibraryTabsContainer.tsx
```diff
@@ -24,25 +24,8 @@ import LibraryTabsQuestions from './LibraryTabsQuestions';
 interface TabContainerProps {
   gameQuestion: GameQuestionType;
   screenSize: ScreenSize;
-  isTabsOpen: boolean;
```
drewjhart, line 27: goodbye prop drilling

central_v2/src/components/librarytabs/LibraryTabsGames.tsx
```diff
@@ -53,66 +40,57 @@ interface LibraryTabsGamesProps<T extends IGameTemplate> {
...
+const isSearchResults = centralData.searchTerms.length > 0;
 const [openTab, setOpenTab] = React.useState(0);
-const handleChange = (event: React.SyntheticEvent, newValue: number) => {
-  if (newValue === 3) {
-    getFav(userProfile);
-  } else if (newValue === 2) {
-    getDrafts();
-  } else {
-    handlePublicPrivateChange(newValue === 1 ? PublicPrivateType.PRIVATE : PublicPrivateType.PUBLIC);
+const [hasInitialized, setHasInitialized] = useState(false);
```
drewjhart, line 64: this code (down to line 71) fetches Elements only on initial run of the code.

central_v2/src/hooks/context/useCentralDataContext.tsx
```diff
@@ -0,0 +1,16 @@
+import { useContext, Dispatch, Context } from 'react';
```
drewjhart, line 1: Custom hooks that will let us use the `centralData` and `centralDataDispatch` that was shown above

central_v2/src/hooks/useCentralDataActions.tsx
```diff
@@ -0,0 +1,515 @@
...
+      direction: SortDirection | null;
+    }
+  ) => void;
+  handleSearchChange: (searchString: string) => void;
+  handlePublicPrivateChange: (newPublicPrivate: PublicPrivateType ) => void;
+  getFav: (user: IUserProfile) => void;
+  getDrafts: () => void;
+  loadMore: () => void;
+}
+
```
drewjhart, line 40: These are all handler functions that connect async API calls with the `CentralData` state in the app. This connects the backend API calls with our context-held state variables and extracts all that logic in a custom hook.

central_v2/src/lib/CentralModels.tsx
```diff
@@ -64,3 +64,16 @@ export type CreateQuestionLocalData = {
   incompleteCards?: IncorrectCard[] | null,
   completeCards?: IncorrectCard[] | null
 }
+
```
drewjhart, line 67: different types of fetch based on the screen/library tab the user has navigated to

central_v2/src/lib/context/CentralDataContext.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { createContext, useReducer } from 'react';
+import { SortDirection, SortType, PublicPrivateType } from '@righton/networking';
+import { centralDataReducer, CentralDataAction } from '../reducer/CentralDataReducer';
+import { ICentralDataState } from './ICentralDataState';
+import { UserStatusType } from '../CentralModels';
+
+// initialize centralDataState
```
drewjhart, line 7: this inits all stateful variables that we'll need to handle across central

central_v2/src/lib/reducer/CentralDataReducer.tsx
```diff
@@ -0,0 +1,75 @@
+import { IGameTemplate, IQuestionTemplate, IUserProfile, GradeTarget, PublicPrivateType, SortType, SortDirection } from '@righton/networking';
+import { ICentralDataState } from '../context/ICentralDataState';
+import { UserStatusType } from '../CentralModels';
+
```
drewjhart, line 4: this represents all actions that can be performed by the reducer to update the CentralData state.

For some of the more complicated updates in useCentralDataActions, I should probably define separate actions here and update them all at once, instead of having multiple calls to the reducer

central_v2/src/components/librarytabs/LibraryTabsGames.tsx
CCosmoe: Gotcha and these are the tab games that we see once we go on the main page right and it populates it with all the games?

central_v2/src/containers/AuthGuard.tsx
```diff
@@ -0,0 +1,24 @@
...
+export default function AuthGuard ({ children }: AuthGuardProps){
+  const isLibrary = useMatch('/library');
+  const { userStatus } = useCentralDataState();
+  console.log(userStatus)
+  
+  if (userStatus === UserStatusType.INCOMPLETE) {
+    return <Navigate to="/nextstep" replace />;
+  }
+
+  if (isLibrary && userStatus === UserStatusType.LOGGEDOUT) {
```
CCosmoe, line 19: Which edge is this one?

central_v2/src/hooks/context/useCentralDataContext.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { CentralDataStateContext, CentralDataDispatchContext } from '../../lib/context/CentralDataContext';
+
+// Custom hook to access state
+export const useCentralDataState = () => {
```
CCosmoe, line 5: Yeah that makes sense thats how its done in the documentation. It will allow us just use the useCentralDataState function rather than importing context everytime and using useContext.

central_v2/src/lib/context/CentralDataContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+const initCentralDataState: ICentralDataState = {
+  userProfile: {
+    title: 'Title...',
+    firstName: '',
+    lastName: '',
+    username: '',
+    email: '',
+    password: '',
+  },
+  userStatus: UserStatusType.LOGGEDOUT,
```
CCosmoe, line 17: Is this saying that if the user status changes to log out you initialize all those states to null or false because the user shouldn't have access to any of that information since no one is logged in?

central_v2/src/hooks/useCentralDataActions.tsx
CCosmoe: Gotcha so since we removed the Auth listener in the back we are doing the validate check inside the userEffect correct?

central_v2/src/lib/reducer/CentralDataReducer.tsx
CCosmoe: Are you thinking about putting that complicated action under one type and then calling dispatch to update those states?

central_v2/src/components/librarytabs/LibraryTabsGames.tsx
drewjhart: Actually, these are just the games that are displayed in the My Library component (rather than the main page elements)

central_v2/src/hooks/useCentralDataActions.tsx
drewjhart: Yeah, so we are:

1. Manually updating the auth status during the sign up process (instead of relying the auth listener)
2. Then we are validating based on that useEffect

central_v2/src/lib/reducer/CentralDataReducer.tsx
drewjhart: Yeah, for an example, I'm thinking instead of this:
```
centralDataDispatch({ type: 'SET_FAV_GAMES', payload: response.games });
centralDataDispatch({ type: 'SET_NEXT_TOKEN', payload: response.nextToken });
centralDataDispatch({ type: 'SET_IS_LOADING', payload: false });
```
we could do something like this:
```
centralDataDispatch({type: 'SET_FAV_GAMES_LOADED', payload: {response.games, response.nextToken, false})
```

PR #1531, opened 2025-03-26
[Central_v2] - Auth Guard

drewjhart:


central_v2/src/components/Header.tsx
```diff
@@ -300,6 +301,7 @@ export default function Header({
                   My Library
                 </ButtonText>
               </TransparentButton>
```
drewjhart, line 303: hey this looks bad

central_v2/src/switches/AppSwitch.tsx
```diff
@@ -52,31 +53,35 @@ function AppSwitch() {
     case questionScreen: {
       currentScreen = ScreenType.QUESTIONS;
       screenComponent = (
-        <ExploreQuestions
```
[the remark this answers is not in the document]
CCosmoe: Dont we need to check the userStatus on login and signup to restrict them from signing up or logging in since the user is either half way signing up or they are logged in?

central_v2/src/components/Header.tsx
CCosmoe: Oh this was the practice comment when you were showing me during our programming session. I started looking for what was wrong here hahaha.

central_v2/src/switches/AppSwitch.tsx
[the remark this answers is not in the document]
drewjhart: yup, you bet. I think we need to further figure out the `userStatus`

central_v2/src/components/Header.tsx
drewjhart: hahahaha yeah sorry, I was being an idiot here hahahaha

PR #1600, opened 2025-05-02
functionlity for userProfile page finished.

CCosmoe:
**Issue:**

Task:
- User should be able to edit their information:
-  ChangeAvatar
-  Username
-  FirstName,
-  LastName,
-  FrontImage
-  BackImage
-  Password

**Resolution:**

Users have given the ability to change their avatar, firstname, lastname, frontimage, and backimage. Giving users the ability to change password has not been implemented yet. Need more information from the design team on figma. 

Relevant code:
- UserProfile.tsx: This component captures the whole page. This is where the UserInterface is laid out for that page.
- UserProfileImageUploadModal.tsx: The component is responsible for letting user upload their image or choose an avatar.
- AuthAPIClient.tsx: A file that lives in networking responsible for connecting with the backend. This is also where we added the function updateCognitoUsername. Function is responsible for updating username in cognito.
- CentralDatamangerAPIClients.tsx: This builds longer functionalities using the functions provided in AuthAPIClient.tsx. This is also where the whole functionality of userProfile is created. 

On the left side of the profile card the game info like gamesmade, questionmade, gamesused don't display data dynamically yet. And the ability to change password is still not implemented as well.

**Preview:**
image
image

### DIFF
--- a/networking/src/APIClients/auth/AuthAPIClient.ts
+++ b/networking/src/APIClients/auth/AuthAPIClient.ts
@@ -17,7 +17,8 @@ import {
   ResendSignUpCodeOutput,
   ConfirmSignUpOutput,
   AuthSession,
-  decodeJWT
+  decodeJWT,
+  updateUserAttributes
 } from 'aws-amplify/auth';
 import { uploadData, downloadData } from 'aws-amplify/storage';
 import amplifyconfig from "../../amplifyconfiguration.json";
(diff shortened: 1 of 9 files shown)


networking/src/APIClients/auth/AuthAPIClient.ts
```diff
@@ -135,7 +136,20 @@ export class AuthAPIClient
       }
   }
   
-  
+  async updateCognitoUsername(newUsername: string): Promise<void> {
```
drewjhart, line 139: nice

networking/src/APIClients/datamanagers/CentralDataManagerAPIClient.ts
```diff
@@ -283,28 +294,45 @@ export class CentralDataManagerAPIClient implements ICentralDataManagerAPIClient
...
     }
-
-    try {
-      if(newProfilePic){
-        const images = await Promise.all([
-          this.authAPIClient.awsUploadImagePrivate(newProfilePic) as any,
-        ]);
-        updatedUser = { ...updatedUser, profilePicPath: images[0].path};
-        console.log("After returning from s3!!", updatedUser)
+    if(frontImage || backImage){
```
drewjhart, line 303: If we have some extra time after our internal QA this week, I bet we can write these so that they use Promise.all and run the authAPIClient calls concurrently (otherwise, these are going to await in series and take longer). Def just focus on Dev QA for now though

Issue #1640, opened 2025-05-13
[Central_v2] - Bug Report - c6 - Priority - P1


chrvtran:
**Describe the bug**
While creating a new account, successfully inputting information and moving on to "Next Steps" to verification and then closing the verification without verifying will lock you out of re-attempting to create an account with that same email with an "Error Signing Up" pop-up message.

Potential reasons for not verifying: (1) no immediate access to email (2) bad internet connection (3) accidental exit

**To Reproduce**
1. From Central, click on "Sign Up"
2. Fill in input areas (remember the email specifically)
3. Click "Next Step"
4. On the Verification screen, close the tab (DO NOT VERIFY)
5. Revisit Central and repeat steps 1-3 (using the same email)
6. You will get an error message pop-up.

**Expected behavior**
If I didn't verify my email, I should be able to make a new account using that email still. Otherwise, I need to be notified of the next steps or wait period. OR the verification email has a link to directly verify.

**Screenshots**

<img width="465" alt="Image" src="https://github.com/user-attachments/assets/331dfed7-1fc2-4095-8639-10acf89304ee" />

**Desktop (please complete the following information):**
 - OS: MacOS Sequora
 - Browser Chrome

**Additional context**
I am 100% sure I did not complete the verification. I am sure on my end that the username is unique. I reproduced the issue with an entirely new email as well, locking out 2 of my emails.

CCosmoe:
@drewjhart isn't this related to 
https://github.com/rightoneducation/righton-app/issues/1545

drewjhart:
Yeah @CCosmoe, I agree. I also think that the behaviour that Chris is flagging is something we'll need to account for. 

So the problem here is that users could be trying to create an account while already having a partially created "unverified account". 

@rod-hinn,  what's the status on #1545 because I do think that impacts how we ultimately will solve this on the `dev` side?

rod-hinn:
Sorry, I forgot to copy and paste this specific solution here, I made a comment on #1541 - do these solutions work?

**Edge Case 3:** If the user signups and doesn’t complete signups full process should they be able to go to other pages?

No—If they log in, it should send them to the verification page to make sure they complete the verification. If they click on any of the pages that only fully verified users have access to, they get redirected to the verification page. But they can browse games and questions.

User Flow

User clicks on sign up
User fills in information
User clicks on next step
User somehow logs out.
User can log back in
User gets redirected to the complete verification page.
User completes verification.
[...]

PR #1706, opened 2025-06-03
[Central_v2] - Bug Resolution - c6

CCosmoe:
Resolves: https://github.com/rightoneducation/righton-app/issues/1640

- At the moment if user is unverified and they try to login, cognito doesn't really send us why they weren't able to login. The amplify sign in function isn't really specific with its errors.
- So to work around this we are relying on fetchUserAttributes and getCurrentUser that will throw an error if the amplify sign in function fails to sign in the user. Another approach we can probably add is maybe adding a lambda function that triggers when user logs in. The lambda function is basically going to check for verification of the email that the user is logging in with.


CCosmoe:
- Need to add resend function. Amplify builtin function.

PR #2591, opened 2026-07-24
[MicroCoach] - Pipeline collaboration docs

drewjhart:


microcoach/docs/prompts/call-3-activity-planning.md
```diff
@@ -0,0 +1,62 @@
...
+A teacher will see activities for [N] misconception(s) in a single session.
+Grade: [grade] | Subject: [subject] | Class size: [cohortSize]
+
+Your job: assign a diverse set of activity structures across all misconceptions so that no two
+misconceptions use the same structure, and the overall session feels varied and pedagogically rich.
+
+Each misconception needs two structure assignments — one for a whole_class activity and one for a
+split_class activity. The two structures for a single misconception must also differ from each other.
+
+## Example Structures (for inspiration — you are not limited to this list)
```
drewjhart, line 31: Here are the activity options that we are currently providing, @sincwu. In terms of promptings (as you were thinking) this is a loose guideline in this prompt, but of course could be tightened to a list of exclusive options for the LLM to choose from.

@dinaragataullina

Issue #2674, opened 2026-09-03
[MicroCoach_v2] - API - Subtask 1 - Align v1 and v2 data structures


drewjhart:
Subtask 1 involves taking the v1 and v2 data structures, analyzing them, and building out the updated data structures that we'll need to run the app with real data. 

This approximately breaks down as:

1. Review the mock data used for MicroCoach_v2. You can find it here: [Mock Data in V2](https://github.com/rightoneducation/righton-app/blob/dev/microcoach_v2/src/lib/mocks/mockPipelineOutput.json)
2. Review the `schema.graphql` from MicroCoach_v1 to see how we did this in v1. You can find it here: [Schema in V1](https://github.com/rightoneducation/righton-app/blob/dev/microcoach/api/amplify/backend/api/microcoach/schema.graphql)
3. Review the current `schema.graphql` from MicroCoach_v2 to see what we are currently doing in v2. You can find it here: [Schema in V2](https://github.com/rightoneducation/righton-app/blob/dev/microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql). Note: this is super preliminary.
4. The goal is to update the v2 `graphql.schema` so that it mirrors the general structure of the v1 `graphql.schema` file, while incorporating the updates into v2. 

Task completion:  push PR with updated `graphql.schema`

PR #2683, opened 2026-09-06
[Microcoach V2] - Defining MicroCoach v2 data schema

buildingBuild:
Issue: #2674

**Description**
Updates the MicroCoach v2 GraphQL schema to support the real-data structure needed for the app, using the [v1 schema](https://github.com/rightoneducation/righton-app/blob/dev/microcoach/api/amplify/backend/api/microcoach/schema.graphql) as a reference and the [v2 mock pipeline output](https://github.com/rightoneducation/righton-app/blob/dev/microcoach_v2/src/lib/mocks/mockPipelineOutput.json) as the target shape.

This updates the preliminary [v2 schema](https://github.com/rightoneducation/righton-app/blob/dev/microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql).

**Resolution:**

Added the core MicroCoach v2 data models needed for real app data, including users, classes, students, sessions, assessments, misconceptions, activities, and saved plans. The schema now follows the general structure of the [v1 schema](https://github.com/rightoneducation/righton-app/blob/dev/microcoach/api/amplify/backend/api/microcoach/schema.graphql), while adding fields and enums based on the [v2 mock pipeline output](https://github.com/rightoneducation/righton-app/blob/dev/microcoach_v2/src/lib/mocks/mockPipelineOutput.json).

Relevant code:
- `microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql` - updated v2 GraphQL schema with real-data models, relationships, and enums.
- `microcoach/api/amplify/backend/api/microcoach/schema.graphql` - v1 schema used as the structure reference.
- `microcoach_v2/src/lib/mocks/mockPipelineOutput.json` - v2 mock data used to identify needed fields and enum values.

**Notes / Questions:**

 - What’s the best way to check this schema before merge? I ran the app build and a local GraphQL parse check, but `amplify api gql-compile` needs Amplify/AWS setup.
- Are we good keeping the big nested UI payloads (`studentWork`, `skillContext`, `routine`, `phases`) as `AWSJSON` for now?
- Should saved plan items stay embedded under `SavedPlan`, or become their own model later?
- I changed `teacherName` to `name` since `User` isn’t teacher-only. Any other naming cleanup we want now, or keep the rest explicit?

**Preview:**

Description of a screenshot highlighting either the original condition (or Figma) or the change
N/A

Description of a screen recording highlighting either the original condition (or Figma) or the change
N/A


microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
   MEMBER
 }
 
-# Embedded (non-@model) type — stored as a list on the User item.
-# Promote to a related @model later if classes need their own queries.
-type Class {
-  id: ID!
-  name: String!
-}
-
```
drewjhart: I'm nitpicking but for each of the @model types, let's add "MicroCoach" in front of the name, so it's easy to filter on the console. So "User" becomes "MicroCoachUser" etc

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
-  teacherName: String
-  classes: [Class]
+  name: String
+  classes: [Class] @hasMany(indexName: "byUser", fields: ["id"])
   role: UserRole! @index(name: "byRole", queryField: "usersByRole")
   createdAt: AWSDateTime!
   updatedAt: AWSDateTime!
 }
+
+type Class @model {
```
drewjhart: Can we rename this to "MicroCoachClassroom"? Maybe it's because I've been working on classes all day for the auth API but it might get confusing down the road

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
+  id: ID!
+  userId: ID! @index(name: "byUser")
+
+  sessions: [Session] @hasMany(indexName: "byClass", fields: ["id"])
+  students: [Student] @hasMany(indexName: "byClass", fields: ["id"])
+
+  name: String!
+  grade: Int
+  state: String
```
drewjhart, line 40: Might be useful to capture state but can I get more of your thinking on why?

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
+  schoolYear: String
+}
+
+type Student @model {
```
drewjhart: This will need to be extended as we get further into the data side of things, but fine to leave for now and we can build it out accordingly.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
+  classId: ID! @index(name: "byClass")
+  sessionLabel: String
+  weekLabel: String
+  weekNumber: Int
+  topic: String
+  ccssStandards: [String]
+  status: SessionStatus # DATA_INGESTED | GENERATED | COMPLETED
+  publishStatus: PublishStatus # "DRAFT" | "PUBLISHED"
+  studentWorksAnalyzed: Int
+  studentsWithStrongUnderstanding: Int
```
drewjhart, line 76: ^Do you know how this gets used later? I could see wanting to retain the student ids here so we can render more info than just the count.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
+  evaluationResults: AWSJSON
+
+  createdAt: AWSDateTime!
+  updatedAt: AWSDateTime!
+}
+
+enum AssessmentType {
+  PPQ
+  POST_PPQ
+  QUIZ
```
drewjhart: I don't think we'll need a quiz type here but lmk if you've seen something in the Figma that I'm missing

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
+}
+
+type Assessment @model {
```
drewjhart: Good skeleton here. I think we'll need to extend this but can leave it until the pipeline work is ready to hook up.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
+}
+
+type Misconception @model {
+  id: ID!
+  sessionId: ID! @index(name: "bySession")
+  classId: ID! @index(name: "byClass")
+  activities: [Activity] @hasMany(indexName: "byMisconception", fields: ["id"])
+
+  rank: Int!
+  badge: Badge
```
drewjhart, line 134: I feel like badge and isRecommendedFocus might be serving the same purpose. What do you think though?

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
+  studentsNeedingSupport: Int
+  studentsUnderstood: Int
+  studentsNoResponse: Int
+  totalAnalyzed: Int!
+  supportSummaryLabel: String
+  understoodSummaryLabel: String
+  shortCountLabel: String
+}
+
+enum ActivityType {
```
drewjhart, line 164: We'll need to keep track of this in particular over the next couple weeks as this list is going to change on the design side. No effort right now, just FYI

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
```diff
@@ -3,30 +3,229 @@ enum UserRole {
...
+  level: GroupingLevel!
+  label: String!
+}
+
+enum PlanStatus {
+  SAVED
+  COMPLETED
+}
+
+type SavedPlan @model {
```
drewjhart: Do you think we need this as a separate table, or can we integrate it into one of the existing ones?

drewjhart (changes requested):
Nice job on this! Very fast turn around and appreciate the thought that has gone into this.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
buildingBuild: I think state might be useful because different states could use different learning standards and its possible microcoach might need that context later down the line as it evolves.  We also wont be able to infer state from the ccss code.  What do you think ?

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
buildingBuild: I added comments above the models we might need to extend since that’ll likely come up in the future. I’m not sure if that matches the style of the codebase though, so let me know if I should remove them

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
buildingBuild: Agree. Ui can derive isRecommendedFocus from badge.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
buildingBuild: Yeah, "QUIZ"  was ported over from the v1 "AssessmentType", but I don’t think it’s actually used.
 Removed it.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
buildingBuild: Absolutely agree. Sinclair talked about this a bit in the bi-weekly meeting too. Just having the count probably won’t be enough. We’ll likely want the student ids so we can show who understood it well and add more detail later. I also added a new field  "studentIdsNeedingSupport". I imagine that would also be useful.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
buildingBuild: Hmm, it depends. We did talk about this a bit. Looking at it more, I think my main clarifying questions are whether SavedPlans is teacher-wide or if each classroom has its own plan, and whether we’ll be tracking individual saved activities over time.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
drewjhart: Hmm, yeah it's a good call. I think we'll need to make sure that we can grab that from the data ingest but having it in the schema and leaving it optional is a good idea. Let's keep it.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
drewjhart: Good call, seems like a good temporary measure. We can pull the comments later down the road.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
drewjhart: Nice, good thinking here.

microcoach_v2/amplify/backend/api/microcoachv2/schema.graphql
drewjhart: Good call:
1. SavedPlans are going to be on a classroom-by-classroom basis.
2. We don't have plans to right now but I could definitely see it in the future

drewjhart: I think if we can decide this one, we can then merge this PR. LMK what you think, I think either keeping it separate or integrating it under a different model is fine.

buildingBuild: Okay I think we can keep it seperate since its possible we could be tracking individual saved activities over time. Ill merge now

PR #2695, opened 2026-09-14
[MicroCoach_v2] - API Foundation

### LINKED ISSUE #2685: [MicroCoach_v2] - API - Subtask 2 - API Foundation
This task consists of building out the API structure so that we can begin to make calls to the backend that you designed and deployed in Subtask 1.

### Current API Structure

The main reference for this work lives in the `networking` folder ([here](https://github.com/rightoneducation/righton-app/tree/dev/networking/src/APIClients)). We will have a very similar setup to this code, as we are performing some pretty similar tasks (namely using GraphQL queries to our backend). Our existing APi structure works as follows:

All API clients are written as classes. The goal here is to be able to write something like this in our react code: 

```
apiclients.microcoachuser.adduser(id);
```

Where all of the more complicated API functionality gets abstracted away in this class structure.

### APIClient Structure

APIClient is the wrapping client that collects all the sub api clients. We do this so, one, we can sequence the spinning up of auth, and aws config in the right order, and two, it makes all other api clients accessible via the `apiclients.` nomenclature above.

### Dedicated APIClients for each data layer

From this main class, you can see subfolders for each of the layers of data that needs to perform its own API logic. Each of these folders includes the APIClient.ts code itself and the associated interface in an interface folder. A good example of this is the gamesession folder [here](https://github.com/rightoneducation/righton-app/blob/dev/networking/src/APIClients/gamesession/TeamAnswerAPIClient.ts).

From here, each APIClient is basically a set of functions to that will intake the required input from the frontend, pair them with the respective autogenerated GraphQL query that is created when `amplify codegen` is run, and then us the AWS functions to send them over to the backend. For an example of it, take a look at the [TeamAnswerAPIClient.ts](https://github.com/rightoneducation/righton-app/blob/dev/networking/src/APIClients/gamesession/TeamAnswerAPIClient.ts).

### Parsing and Models

The final complicated factor is that AWS is going to store data on the backend in formats that will differ than what we need on the frontend. For example, AWSJSON is stored as a string on the backend, so we'll need to parse that out using JSON.parse.

As such, there is also a parser for each of these APIClients that handles the movement of data between our app (using our interfaces) and the AWS side (using their interfaces). The parser will include interfaces for both sides of the data, as well as functions that convert in both directions. You can find the parsers in the [Parsers](https://github.com/rightoneducation/righton-app/tree/dev/networking/src/Parsers) folder and the interfaces in the [Models ](https://github.com/rightoneducation/righton-app/tree/dev/networking/src/Models)folder.

I've started this work already in MicroCoach_v2 as a build up the teacher signup and login flows. So you'll see the above folder structure roughly sketched out in your branch (I synched the two up so you just have to git pull).

I think this should give you enough to get started but let me know if you have questions!

buildingBuild:
**Issue:**

Resolves #2685.

Builds out the MicroCoach v2 API client foundation 

**Resolution:**

Added api models, parsers, client interfaces, and client classes for the MicroCoach v2 data layers. They wrap the Amplify-generated GraphQL queries/mutations from `AWSAPI.ts`, `graphql/queries.ts`, and `graphql/mutations.ts`, then parse AppSync response shapes into frontend-facing models.

Also updated the existing user API client to use the current `MicroCoachUser` generated types and operations.

Relevant code:
- `src/api/APIClients/APIClients.ts` - registers the MicroCoach API clients on the shared wrapper.
- `src/api/APIClients/interfaces/IAPIClients.ts` - exposes the typed API client contract.
- `src/api/APIClients/*/*APIClient.ts` - adds class-based API clients for MicroCoach models.
- `src/api/APIClients/*/interfaces/*APIClient.ts` - defines frontend-facing API client method contracts.
- `src/api/Parsers/*Parser.ts` - maps AWS/AppSync response shapes into frontend models.
- `src/api/Models/AWS/*` - defines AWS/AppSync response model shapes.
- `src/api/Models/IMicroCoach*.ts` - defines frontend-facing API models for MicroCoach data.

<details>
<summary> My thought process while working on the task </summary>
Hi Drew, this was basically my mental model while doing the task and trying to understand some of it. A lot of these are things you already mentioned or things that were mentioned in the task, but I wanted to put it into text so you could see where my head was at while I was working through it. So if some things are repeated, that’s why.

So basically, this is how I understand the MicroCoach v2 API setup right now. I’m mostly just talking through how I think all of these files work together, because I spent a lot of time trying to map out what was actually happening in the codebase.

Starting with the schema, we have our models defined inside schema.graphql. Whenever that schema changes, we run Amplify codegen, and that generates the AWS/AppSync types and GraphQL operations that the frontend can actually use. So files like AWSAPI.ts, and the generated queries and mutations, are basically telling us what the backend schema currently supports and what types of inputs and outputs we should expect.

If the schema changes, codegen can regenerate them anyway, and if something is stale there, it probably means the generated code just needs to be updated from the schema.

Then we have the parsers. The way I understand the parsers is basically that AWS/AppSync gives us data in whatever generated shape it has, and then the parser converts that into the shape we actually want the frontend to work with (Users model to MicroCoachUsers, for example).

So, for example, if AWS gives us something like an AWSJSON field as a JSON string, the parser could take that, parse it, and return an actual JavaScript object instead. Or generally, if there’s anything about the generated AWS response that we don’t want leaking directly into the rest of the frontend, the parser is kind of that conversion layer.

Then we move into the actual API clients.

For each of the main models, we basically have an API client folder. Inside that folder there’s usually an interface and then the actual API client implementation.

The interface file, which is usually prefixed with I, is basically saying, “These are the methods that this API client should expose.” So it might say this client has something like createUser, getUser, updateUser, whatever methods make sense for that model.

Then the actual APIClient class is where those methods are implemented. That’s the part that is actually talking to the GraphQL backend. So that class would use the generated GraphQL operations, make the request through the existing base API client setup, get the response back, and then usually pass that response through the parser before returning it.

So I kind of think of the flow there as:

frontend calls API client method → API client makes GraphQL request → AppSync/AWS responds → parser converts the response → frontend gets the cleaner model.

What I struggled with was deciding what methods specifically to implement. I looked through schema.graphql for clues and then used the generated AWS files to give me an idea of what methods made sense for each API client.

For example, if I see something like CreateMicroCoachUserInput generated from the schema, then that tells me there’s obviously a create operation available for MicroCoachUser, so the client would probably expose something like createUser, and that method would take the generated CreateMicroCoachUserInput type.

Same thing with IDs. I used the schema to help figure out whether a method should use something like an id or classId.

So that’s kind of how I was deciding what methods belonged inside each API client instead of just making random CRUD methods.

Then there are the two higher-level files, IAPIClients and APIClients.

My understanding of IAPIClients is that it defines the overall shape of our API client collection. Basically, it says that our shared API client object should have things like a user client, activity client, assessment client, misconception client, and so on.

Then the actual APIClients class implements that interface.

The constructor is basically wiring all of those individual API clients together when an APIClients instance is created. So instead of React components having to instantiate all of these different clients themselves, they can use that shared wrapper.

And I think that’s ultimately what gives us the nice frontend usage where we can do something like:

apiClients.microCoachUser.createUser(…)

If my mental model here is correct, I feel like I have a way stronger understanding of the repo and how the system works now than I did before. Fingers crossed 🤞

</details>


microcoach_v2/src/api/APIClients/activity/ActivityAPIClient.ts
```diff
@@ -0,0 +1,33 @@
...
+import { IActivity } from "../../../lib/PipelineModels";
+import { ActivityParser } from "../../Parsers/ActivityParser";
+import { BaseAPIClient, GraphQLOptions } from "../base/BaseAPIClient";
+import { IActivityAPIClient } from "./interfaces/IActivityAPIClient";
+
+export class ActivityAPIClient
+  extends BaseAPIClient
+  implements IActivityAPIClient
+{
+  async getActivitiesByMisconceptionId(misconceptionId: string): Promise<IActivity[]> {
```
drewjhart: I think there's something odd happening with "Activity" vs "Activities" here. I think Activity will probably need create, update and delete functions and we can probably just use singular activity. 

I think the issue might be in our `schema.graphql` here:
```
type MicroCoachMisconception @model @auth(rules: [{ allow: private }]) {
  id: ID!
  sessionId: ID! @index(name: "bySession")
  classId: ID! @index(name: "byClass")
  activities: [MicroCoachActivity]
    @hasMany(indexName: "byMisconception", fields: ["id"])
```

(ln 131)

The has many might be the thing that is creating the 'activities' generated queries. I thin if we call this activity and codegen, and then have MicroCoachActivity as the basis for all our queries here. LMK if this doesn't fix it though.

microcoach_v2/src/api/APIClients/session/SessionAPIClient.ts
```diff
@@ -0,0 +1,50 @@
...
+import { IMicroCoachSession } from "../../Models/IMicroCoachSession";
+import { SessionParser } from "../../Parsers/SessionParser";
+import { BaseAPIClient, GraphQLOptions } from "../base/BaseAPIClient";
+import { ISessionAPIClient } from "./interfaces/ISessionAPIClient";
+
+export class SessionAPIClient
+  extends BaseAPIClient
+  implements ISessionAPIClient
+{
+  async getSession(id: string): Promise<IMicroCoachSession | null> {
```
drewjhart, line 31: Might as well add create, update and delete on here.

microcoach_v2/src/api/Parsers/ActivityParser.ts
```diff
@@ -0,0 +1,43 @@
+import { AWSActivity } from "../Models/AWS/AWSActivity";
+import {
+  IActivity,
+  IActivityPhases,
+  IRoutine,
+} from "../../lib/PipelineModels";
+
+export class ActivityParser {
```
drewjhart, line 6: On any of these parsers for elements that include mutations (changes to the backend) in the form of create, update, delete etc, we'll need to go the opposite way (AWSActivity from IActivity) for example, so that we can pass the AWSActivity as a param to the mutation call

microcoach_v2/src/api/Parsers/ActivityParser.ts
```diff
@@ -0,0 +1,43 @@
...
+          }
+        : null,
+      targets: activity.targets ?? null,
+      instructionalMove: activity.instructionalMove ?? null,
+      strategyTag: activity.strategyTag ?? null,
+      phases: activity.phases
+        ? JSON.parse(activity.phases) as IActivityPhases
+        : null,
+    }
+
```
drewjhart, line 55: I think if we have any fields that are ! on the AWS side, we should have a thrown Error condition here on a null on that field. You can see precedent for that here:

https://github.com/rightoneducation/righton-app/blob/b60aa29f4f76b3dde6dad1428ff1c1ca4fa3b3a0/networking/src/Parsers/TeamAnswerParser.ts#L78

microcoach_v2/src/api/Models/IMicroCoachAssessment.ts
```diff
@@ -0,0 +1,12 @@
+import { AssessmentType } from "../../AWSAPI";
```
drewjhart, line 1: This is nitpicky but can we move IMicroCoachActivity, IMicroCoachMisconception out of pipeline models and into separate ts files here?

microcoach_v2/src/api/Parsers/MisconceptionParser.ts
```diff
@@ -0,0 +1,42 @@
...
+      badge: misconception.badge ?? null,
+      isRecommendedFocus: misconception.badge === "RECOMMENDED_FOCUS",
+      title: misconception.title,
+      titleCased: misconception.titleCased ?? misconception.title,
+      shortLabel: misconception.shortLabel ?? misconception.title,
+      description: misconception.description,
+      consequence: misconception.consequence ?? "",
+
+      prevalence: {
+        level: misconception.prevalence?.level ?? "FEW", // would this be a good fallback ?
```
drewjhart: For now, I think that makes sense for sure but I think we can import the enum here and use that via PrevalenceLevel.FEW. I think this applies to any reference in the parsers that tie to enums (like the RECOMMENDED_FOCUS) above

drewjhart (changes requested):

drewjhart:
@buildingBuild Nice work on this, really solid effort and scale across the different layers. For your mental model - yup bang on! That covers the structure really closely. Got it!

microcoach_v2/src/api/Parsers/ActivityParser.ts
buildingBuild: Ahhh I see, so for anything the frontend might make edits to, we’d need to handle the conversion back to the aws type as well. That makes sense. Will make the corrections

